PR #1, opened 2021-05-24
Add azure IoT Sample

hihigupt:
Add azure IoT Sample

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libs/azure-iot-middleware-freertos"]
+	path = libs/azure-iot-middleware-freertos
+	url = https://github.com/Azure/azure-iot-middleware-freertos.git
--- a/demos/sample_azure_iot/sample_azure_iot.c
+++ b/demos/sample_azure_iot/sample_azure_iot.c
@@ -1,0 +1,628 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+/* Standard includes. */
+#include <string.h>
+#include <stdio.h>
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "task.h"
+
+/* Demo Specific configs. */
+#include "demo_config.h"
+
+/* Azure Provisioning/IoT Hub library includes */
+#include "azure_iot_hub_client.h"
+#include "azure_iot_provisioning_client.h"
+
+/* Exponential backoff retry include. */
+#include "backoff_algorithm.h"
+
+/* Transport interface implementation include header for TLS. */
+#include "transport_tls_socket.h"
+
+/* Crypto helper header. */
+#include "crypto.h"
+/*-----------------------------------------------------------*/
+
+/* Compile time error for undefined configs. */
+#if !defined( democonfigHOSTNAME ) && !defined( democonfigENABLE_DPS_SAMPLE )
+    #error "Define the config democonfigHOSTNAME by following the instructions in file demo_config.h."
+#endif
+
+#if !defined( democonfigENDPOINT ) && defined( democonfigENABLE_DPS_SAMPLE )
+    #error "Define the config dps endpoint by following the instructions in file demo_config.h."
+#endif
+
+#ifndef democonfigROOT_CA_PEM
+    #error "Please define Root CA certificate of the IoT Hub(democonfigROOT_CA_PEM) in demo_config.h."
+#endif
+
+#if defined( democonfigDEVICE_SYMMETRIC_KEY ) && defined( democonfigCLIENT_CERTIFICATE_PEM )
+    #error "Please define only one auth democonfigDEVICE_SYMMETRIC_KEY or democonfigCLIENT_CERTIFICATE_PEM in demo_config.h."
+#endif
+
+#if !defined( democonfigDEVICE_SYMMETRIC_KEY ) && !defined( democonfigCLIENT_CERTIFICATE_PEM )
+    #error "Please define one auth democonfigDEVICE_SYMMETRIC_KEY or democonfigCLIENT_CERTIFICATE_PEM in demo_config.h."
+#endif
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief The maximum number of retries for network operation with server.
+ */
+#define sampleazureiotRETRY_MAX_ATTEMPTS                      ( 5U )
+
+/**
+ * @brief The maximum back-off delay (in milliseconds) for retrying failed operation
+ *  with server.
+ */
+#define sampleazureiotRETRY_MAX_BACKOFF_DELAY_MS              ( 5000U )
+
+/**
+ * @brief The base back-off delay (in milliseconds) to use for network operation retry
+ * attempts.
+ */
+#define sampleazureiotRETRY_BACKOFF_BASE_MS                   ( 500U )
+
+/**
+ * @brief Timeout for receiving CONNACK packet in milliseconds.
+ */
+#define sampleazureiotCONNACK_RECV_TIMEOUT_MS                 ( 10 * 1000U )
+
+/**
+ * @brief The Telemetry message published in this example.
+ */
+#define sampleazureiotMESSAGE                                 "Hello World : %lu !"
+
+/**
+ * @brief The reported property payload to send to IoT Hub
+ */
+#define sampleazureiotTWIN_PROPERTY                           "{ \"TwinIterationForCurrentConnection\": \"%lu\" }"
+
+/**
+ * @brief Time in ticks to wait between each cycle of the demo implemented
+ * by prvMQTTDemoTask().
+ */
+#define sampleazureiotDELAY_BETWEEN_DEMO_ITERATIONS_TICKS     ( pdMS_TO_TICKS( 5000U ) )
+
+/**
+ * @brief Timeout for MQTT_ProcessLoop in milliseconds.
+ */
+#define sampleazureiotPROCESS_LOOP_TIMEOUT_MS                 ( 500U )
+
+/**
+ * @brief Delay (in ticks) between consecutive cycles of MQTT publish operations in a
+ * demo iteration.
+ *
+ * Note that the process loop also has a timeout, so the total time between
+ * publishes is the sum of the two delays.
+ */
+#define sampleazureiotDELAY_BETWEEN_PUBLISHES_TICKS           ( pdMS_TO_TICKS( 2000U ) )
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define sampleazureiotTRANSPORT_SEND_RECV_TIMEOUT_MS          ( 2000U )
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define sampleazureiotProvisioning_Registration_TIMEOUT_MS    ( 20U )
+
+/**
+ * @brief Wait timeout for subscribe to finish.
+ */
+#define sampleazureiotSUBSCRIBE_TIMEOUT                       ( 10 * 1000U )
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Unix time.
+ *
+ * @return Time in milliseconds.
+ */
+uint64_t ullGetUnixTime( void );
+/*-----------------------------------------------------------*/
+
+/* Define buffer for IoT Hub info.  */
+#ifdef democonfigENABLE_DPS_SAMPLE
+    static uint8_t ucSampleIotHubHostname[ 128 ];
+    static uint8_t ucSampleIotHubDeviceId[ 128 ];
+    static AzureIoTProvisioningClient_t xAzureIoTProvisioningClient;
+#endif /* democonfigENABLE_DPS_SAMPLE */
+
+static uint8_t ucPropertyBuffer[ 32 ];
+static uint8_t ucScratchBuffer[ 128 ];
+
+/* Each compilation unit must define the NetworkContext struct. */
+struct NetworkContext
+{
+    TlsTransportParams_t * pParams;
+};
+
+static AzureIoTHubClient_t xAzureIoTHubClient;
+/*-----------------------------------------------------------*/
+
+#ifdef democonfigENABLE_DPS_SAMPLE
+
+/**
+ * @brief Gets the IoT Hub endpoint and deviceId from Provisioning service.
+ *   This function will block for Provisioning service for result or return failure.
+ * 
+ * @param[in] pXNetworkCredentials  Network credential used to connect to Provisioning service
+ * @param[out] ppucIothubHostname  Pointer to uint8_t* IoT Hub hostname return from Provisioning Service
+ * @param[in,out] pulIothubHostnameLength  Length of hostname
+ * @param[out] ppucIothubDeviceId  Pointer to uint8_t* deviceId return from Provisioning Service
+ * @param[in,out] pulIothubDeviceIdLength  Length of deviceId
+ */
+    static uint32_t prvIoTHubInfoGet( NetworkCredentials_t * pXNetworkCredentials,
+                                      uint8_t ** ppucIothubHostname,
+                                      uint32_t * pulIothubHostnameLength,
+                                      uint8_t ** ppucIothubDeviceId,
+                                      uint32_t * pulIothubDeviceIdLength );
+
+#endif /* democonfigENABLE_DPS_SAMPLE */
+
+/**
+ * @brief The task used to demonstrate the MQTT API.
+ *
+ * @param[in] pvParameters Parameters as passed at the time of task creation. Not
+ * used in this example.
+ */
+static void prvAzureDemoTask( void * pvParameters );
+
+/**
+ * @brief Connect to IoT Hub with reconnection retries.
+ * 
+ * If connection fails, retry is attempted after a timeout.
+ * Timeout value will exponentially increase until maximum
+ * timeout value is reached or the number of attempts are exhausted.
+ * 
+ * @param pcHostName Hostname of the endpoint to connect to.
+ * @param ulPort Endpoint port.
+ * @param pxNetworkCredentials Pointer to Network credentials.
+ * @param pxNetworkContext Point to Network context created.
+ * @return uint32_t The status of the final connection attempt.
+ */
+static uint32_t prvConnectToServerWithBackoffRetries( const char * pcHostName,
+                                                      uint32_t ulPort,
+                                                      NetworkCredentials_t * pxNetworkCredentials,
+                                                      NetworkContext_t * pxNetworkContext );
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Static buffer used to hold MQTT messages being sent and received.
+ */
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libs/azure-iot-middleware-freertos"]
+	path = libs/azure-iot-middleware-freertos
+	url = git@github.com:Azure/azure-iot-middleware-freertos.git
```
danewalton, line 3: I don't know if I've ever commented on a `.gitmodules` file so this is a first for me :)

Can we make this the http version of the repo for greater compatibility on pulling it? It could potentially complicate the process for contributing as someone would have to set up their username/password by default but it increases the compatibility if they don't have ssh set up. Plus they can always change the url to the ssh one if they want to contribute and use ssh.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,624 @@
...
+/**
+ * @brief The base back-off delay (in milliseconds) to use for network operation retry
+ * attempts.
+ */
+#define sampleazureiotRETRY_BACKOFF_BASE_MS                   ( 500U )
+
+/**
+ * @brief Timeout for receiving CONNACK packet in milliseconds.
+ */
+#define sampleazureiotCONNACK_RECV_TIMEOUT_MS                 ( 1000U )
```
danewalton, line 73: make this 10 like we have in the e2e?
```suggestion
#define sampleazureiotCONNACK_RECV_TIMEOUT_MS                 ( 10 * 1000U )
```

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,624 @@
...
+/*-----------------------------------------------------------*/
+
+#ifdef democonfigENABLE_DPS_SAMPLE
+
+/**
+ * @brief Gets the IoTHub endpoint and deviceId from Provisioning service.
+ *
+ * @param[in] pXNetworkCredentials  Network credential used to connect to Provisioning service
+ * @param[out] ppucIothubHostname  Pointer to uint8_t* IoTHub hostname return from Provisioning Service
+ * @param[inout] pulIothubHostnameLength  Length of hostname
```
danewalton, line 154: nit here and elsewhere: 
```suggestion
 * @param[in,out] pulIothubHostnameLength  Length of hostname
```

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,624 @@
...
+    #else
+        uint8_t * pucIotHubHostname = ( uint8_t * ) democonfigHOSTNAME;
+        uint8_t * pucIotHubDeviceId = ( uint8_t * ) democonfigDEVICE_ID;
+        uint32_t pulIothubHostnameLength = sizeof( democonfigHOSTNAME ) - 1;
+        uint32_t pulIothubDeviceIdLength = sizeof( democonfigDEVICE_ID ) - 1;
+    #endif /* democonfigENABLE_DPS_SAMPLE */
+
+    ( void ) pvParameters;
+
+    configASSERT( AzureIoT_Init() == eAzureIoTSuccess );
```
danewalton, line 321: nit: add a comment here what this is initializing? logging?

danewalton (approved):

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+/* Transport interface implementation include header for TLS. */
+#include "transport_tls_socket.h"
+
+/* Crypto helper header. */
+#include "crypto.h"
+/*-----------------------------------------------------------*/
+
+/* Compile time error for undefined configs. */
+#if !defined( democonfigHOSTNAME ) && !defined( democonfigENABLE_DPS_SAMPLE )
+    #error "Define the config democonfigHOSTNAME by following the instructions in file demo_config.h."
```
jspaith, line 32: Is there a `demo_config.h` that will be going to come in a separate PR?  If so it's perfectly fine, but if not we should discuss offline how customer would get this.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+    TlsTransportParams_t * pParams;
+};
+
+static AzureIoTHubClient_t xAzureIoTHubClient;
+/*-----------------------------------------------------------*/
+
+#ifdef democonfigENABLE_DPS_SAMPLE
+
+/**
+ * @brief Gets the IoTHub endpoint and deviceId from Provisioning service.
```
jspaith, line 150: Nit - IoT Hub should be two words when in comments.  Here & other places.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+
+
+/**
+ * @brief Connect to MQTT broker with reconnection retries.
+ *
+ * If connection fails, retry is attempted after a timeout.
+ * Timeout value will exponentially increase until maximum
+ * timeout value is reached or the number of attempts are exhausted.
+ *
+ * @param[out] pxNetworkContext The parameter to return the created network context.
```
jspaith, line 182: There's only 1 parameter doc'd in doxygen for this but function has more.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+static TlsTransportStatus_t prvConnectToServerWithBackoffRetries( const char * pcHostName,
+                                                                  uint32_t ulPort,
+                                                                  NetworkCredentials_t * pxNetworkCredentials,
+                                                                  NetworkContext_t * pxNetworkContext );
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Static buffer used to hold MQTT messages being sent and received.
+ */
+static uint8_t ucSharedBuffer[ democonfigNETWORK_BUFFER_SIZE ];
```
jspaith, line 195: If comment is correct and it's used for MQTT messages *only* could we rename to like `ucMqttMessageBuffer` or something more descriptive?  Or if it's used all over the place we should update comment & leave naming as-is.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+
+/**
+ * @brief Global start unix time
+ */
+static uint64_t ulGlobalEntryTime = 1639093301;
```
jspaith, line 200: This variable doesn't seem to be used anywhere.  If it is needed, i'd prefer if we could get from system though not sure if time.h or some embedded PAL would have issue with this or at least with us in a generic way??

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+static void prvHandleDirectMethod( AzureIoTHubClientMethodRequest_t * pxMessage,
+                                   void * pvContext )
+{
+    LogInfo( ( "Method payload : %.*s \r\n",
+               pxMessage->ulPayloadLength,
+               pxMessage->pvMessagePayload ) );
+
+    AzureIoTHubClient_t * xHandle = ( AzureIoTHubClient_t * ) pvContext;
+
+    if( AzureIoTHubClient_SendMethodResponse( xHandle, pxMessage, 200,
```
jspaith, line 229: PART1 - 
@danewalton / @hihigupt  - I just realized `AzureIoTHubClient_SendMethodResponse` in product takes `AzureIoTHubClientMethodRequest_t` as an input parameter.

Not this PR I know, but i think this should be a MethodResponse?  We might add response specific fields at some point, and even today the componentName/commandName I think would be ignored on response.

PART2 - 
It looks like this is echoing back whatever server sent to us?  That seems fair enough, no matter what we do with the MethodResponse i suggest in part1.  But we should clearly comment that that's the intent here.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+        pxNetworkCredentials->pPrivateKey = ( const unsigned char * ) democonfigCLIENT_PRIVATE_KEY_PEM;
+        pxNetworkCredentials->privateKeySize = sizeof( democonfigCLIENT_PRIVATE_KEY_PEM );
+    #endif
+
+    return 0;
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Azure IoT demo task.
```
jspaith, line 290: What is a "demo task"?  I take it this basically also acts as the (de facto) main()?  Feel this could use another line or two about this is the main interaction point or some such?

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+        LogInfo( ( "Short delay before starting the next iteration.... \r\n\r\n" ) );
+        vTaskDelay( sampleazureiotDELAY_BETWEEN_DEMO_ITERATIONS_TICKS );
+    }
+}
+/*-----------------------------------------------------------*/
+
+#ifdef democonfigENABLE_DPS_SAMPLE
+
+    /**
+    * @brief Get IoTHub endpoint and device Id info, when Provisioning service is used.
```
jspaith, line 475: Nit - I think we should doc that this might take "a while" to run - like seconds, presumably configured by some #define macro?

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+         * packet. After sending disconnect, client must close the network
+         * connection. */
+        xResult = AzureIoTHubClient_Disconnect( &xAzureIoTHubClient );
+        configASSERT( xResult == eAzureIoTHubClientSuccess );
+
+        /* Close the network connection.  */
+        TLS_Socket_Disconnect( &xNetworkContext );
+
+        /* Wait for some time between two iterations to ensure that we do not
+         * bombard the broker. */
```
jspaith, line 464: We shouldn't use word "broker" here but instead "IoT Hub" or some such.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+    BackoffAlgorithmContext_t xReconnectParams;
+    uint16_t usNextRetryBackOff = 0U;
+
+    /* Initialize reconnect attempts and interval. */
+    BackoffAlgorithm_InitializeParams( &xReconnectParams,
+                                       sampleazureiotRETRY_BACKOFF_BASE_MS,
+                                       sampleazureiotRETRY_MAX_BACKOFF_DELAY_MS,
+                                       sampleazureiotRETRY_MAX_ATTEMPTS );
+
+    /* Attempt to connect to MQTT broker. If connection fails, retry after
```
jspaith, line 571: Simialr feedback about not using words "MQTT broker" here & other places in this it may come up.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+            /* Generate a random number and calculate backoff value (in milliseconds) for
+             * the next connection retry.
+             * Note: It is recommended to seed the random number generator with a device-specific
+             * entropy source so that possibility of multiple devices retrying failed network operations
+             * at similar intervals can be avoided. */
+            xBackoffAlgStatus = BackoffAlgorithm_GetNextBackoff( &xReconnectParams, configRAND32(), &usNextRetryBackOff );
+
+            if( xBackoffAlgStatus == BackoffAlgorithmRetriesExhausted )
+            {
+                LogError( ( "Connection to the broker failed, all attempts exhausted." ) );
```
jspaith, line 596: Please log the `xNetworkStatus` both here and even if we have additional retries left.  This is very helpful for troubleshooting.

Also when we're in the "Retrying..." log please print how many milliseconds it'll be till next retry.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+        /* Sends an MQTT Connect packet over the already established TLS connection,
+         * and waits for connection acknowledgment (CONNACK) packet. */
+        LogInfo( ( "Creating an MQTT connection to %s.\r\n", pucIotHubHostname ) );
+
+        xResult = AzureIoTHubClient_Connect( &xAzureIoTHubClient,
+                                             false, &xSessionPresent,
+                                             sampleazureiotCONNACK_RECV_TIMEOUT_MS );
+        configASSERT( xResult == eAzureIoTHubClientSuccess );
+
+        xResult = AzureIoTHubClient_SubscribeCloudToDeviceMessage( &xAzureIoTHubClient, prvHandleCloudMessage,
```
jspaith, line 391: Do we really want to block the call for `sampleazureiotWAIT_FOREVER`?  I'm not sure that's great practice since "stuff happens".  We don't block forever on CONNAC, its configurable, i think there could be one sampleblahBlah_SUBSCRIBE_RECV_TIMEOUT for all the subscribers (don't need a per function timeout i don't tihkn).

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+
+        xResult = AzureIoTHubClient_SubscribeDirectMethod( &xAzureIoTHubClient, prvHandleDirectMethod,
+                                                           &xAzureIoTHubClient, sampleazureiotWAIT_FOREVER );
+        configASSERT( xResult == eAzureIoTHubClientSuccess );
+
+        xResult = AzureIoTHubClient_SubscribeDeviceTwin( &xAzureIoTHubClient, prvHandleDeviceTwinMessage,
+                                                         &xAzureIoTHubClient, sampleazureiotWAIT_FOREVER );
+        configASSERT( xResult == eAzureIoTHubClientSuccess );
+
+        /* Get the device twin on boot */
```
jspaith, line 403: Nit "on boot" should probably be "after initial connection" since we don't know when sample would run in practice.

jspaith: @danewalton  @hihigupt  - another API feedback now that I'm staring at samples more.  I think `AzureIoTHubClient_GetDeviceTwin` is over-promising.  Becaues we don't actually get the twin in this call.

It's kinda more like `AzureIoTHubClient_RequestDeviceTwinAsync` (don't know about `Async` bit, in IoT C SDK it'd be 100% required).  ?

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+            LogInfo( ( "Attempt to receive publish message from IoTHub.\r\n" ) );
+            xResult = AzureIoTHubClient_ProcessLoop( &xAzureIoTHubClient,
+                                                     sampleazureiotPROCESS_LOOP_TIMEOUT_MS );
+            configASSERT( xResult == eAzureIoTHubClientSuccess );
+
+            if( ulPublishCount % 2 == 0 )
+            {
+                /* Send reported property every other cycle */
+                xResult = AzureIoTHubClient_SendDeviceTwinReported( &xAzureIoTHubClient,
+                                                                    ( const uint8_t * ) sampleazureiotTWIN_PROPERTY,
```
jspaith, line 433: Consider on this one doing something like 
{

   char sampleazureiotTWIN_PROPERTY[SomeBufferSize];
   sprintf(sampleazureiotTWIN_PROPERTY, "TwinIterationForCurrentConnection:%d", ++someGlobalCounter)"
}

Check that sprintf() is cool on FreeRTOS or potentiall they have a stripped down one, right, that could handle a %d.

The reason I recommend this is that repeatedl telemetry sends I can easily see in IoT Explorer, even if of same data.  But I don't get the sense that the data is being updated by looking at a DeviceTwin if we always send the same thing.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+                                                                  sizeof( democonfigDEVICE_SYMMETRIC_KEY ) - 1,
+                                                                  Crypto_HMAC );
+            configASSERT( xResult == eAzureIoTProvisioningSuccess );
+        #endif // democonfigDEVICE_SYMMETRIC_KEY
+
+        do
+        {
+            xResult = AzureIoTProvisioningClient_Register( &xAzureIoTProvisioningClient,
+                                                           sampleazureiotProvisioning_Registration_TIMEOUT_MS );
+        } while( xResult == eAzureIoTProvisioningPending );
```
jspaith, line 527: Nit: IIRC there's various "limit max number of retries down in the _Register" function.  Or maybe not, if we always get pending??.

In any case I think we sholud have some limit of max number of times we'll go through this loop.  Unfortunately if DPS is down and always pending there's not a ton to be done but it could at least be used a best practice to show a bail-out.  How app would actually handle such a thing - maybe via a watchdog or longer timeout - is way outside scope of this straightforward sample of course.

demos/sample_azure_iot/sample_azure_iot.c
hihigupt: yes, demo_config.h is something that is application specific and will be part of separate PR.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -0,0 +1,625 @@
...
+    ( void ) pvContext;
+
+    LogInfo( ( "Cloud message payload : %.*s \r\n",
+               pxMessage->ulPayloadLength,
+               pxMessage->pvMessagePayload ) );
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Direct method message callback handler
```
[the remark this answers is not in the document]
hihigupt: Will defer this change till Dane gets his PR and add PNP sample

demos/sample_azure_iot/sample_azure_iot.c
hihigupt: how about AzureIoTHubClient_RequestDeviceTwin name?

demos/sample_azure_iot/sample_azure_iot.c
hihigupt: Currently assumption is that state machine in DPS always move to Error state, if any failure happen in any state (like sending or response failure). As MQTT currently sits over TCP and we do MQTT keep alive, we will bail out of this loop as soon as network error/service error occurs. 

Based on above I wanted this logic to be simple. Please let me know if you think otherwise.

demos/sample_azure_iot/sample_azure_iot.c
hihigupt: Updated comment and created issue for naming api
https://github.com/Azure/azure-iot-middleware-freertos/issues/127

demos/sample_azure_iot/sample_azure_iot.c
hihigupt: PART1 : I agree we can revisit if we need the response struct.

PART2: I did not do that because in that case I have to construct JSON response that contains random payload we got in method. To keep it simple I just return empty json.

hihigupt: added details in issue: for now resolving this.

demos/sample_azure_iot/sample_azure_iot.c
jspaith: If the product code does the right thing - and we have it contractually obligated to do so always (which I think back-compat 101 would guarantee?) then agree with keeping sample simple.

PR #4, opened 2021-06-04
Add cmake files for linux port

hihigupt:
Add cmake files for linux port

### DIFF
--- a/.github/scripts/ci_tests.sh
+++ b/.github/scripts/ci_tests.sh
@@ -1,0 +1,47 @@
+#! /bin/bash
+
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+set -o errexit # Exit if command failed.
+set -o nounset # Exit if variable not set.
+set -o pipefail # Exit if pipe failed.
+
+echo -e "::group::FreeRTOS Source"
+TEST_FREERTOS_COMMIT_ID=c8fa483b68c6c1149c2a7a8bc1e901b38860ec9b
+
+if [ ! -d "libs/FreeRTOS" ]; then
+    git clone https://github.com/FreeRTOS/FreeRTOS.git libs/FreeRTOS
+    pushd libs/FreeRTOS
+    git checkout -b ${TEST_FREERTOS_COMMIT_ID}
+    git submodule sync
+    git submodule update --init --recursive --depth=1
+    popd
+fi
+TEST_FREERTOS_SRC=`pwd`/libs/FreeRTOS
+
+function exit_if_binary_does_not_exist()
+{
+    local search_dir=$1
+    local filename=$2
+
+    if [ $(find "$search_dir" -name "$filename") ]; then 
+      echo "$filename is found in $search_dir"
+    else
+      echo "$filename not found"
+      exit 1
+    fi
+}
+
+function sample_build() {
+    local vendor=$1
+    local board=$2
+    local outdir=$3
+
+    cmake -G Ninja -DBOARD=$board -DVENDOR=$vendor -B$outdir -DFREERTOS_PATH=$TEST_FREERTOS_SRC .
+    cmake --build $outdir
+}
+
+echo -e "::group::Building sample for linux port"
+sample_build "PC" "linux" "build_pc_linux"
+exit_if_binary_does_not_exist "build_pc_linux" "iot-middleware-sample.elf"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+cmake_minimum_required(VERSION 3.13)
+set(CMAKE_TOOLCHAIN_FILE cmake/${VENDOR}_gcc.cmake)
+
+project(iot-middleware-sample C ASM)
+set(CMAKE_INCLUDE_CURRENT_DIR TRUE)
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+
+# include common utilities
+include(common/utilities)
+
+# Set FreeRTOS path
+if(NOT FREERTOS_PATH)
+    free_rtos_fetch()
+    message(STATUS "No FREERTOS_PATH specified: Fetching FreeRTOS to path <${FREERTOS_PATH}>")
+endif()
+
+find_package(FreeRTOS REQUIRED)
+find_package(FreeRTOSPlus REQUIRED)
+
+# Include logging globally
+include_directories("${FreeRTOSPlus_PATH}/Source/Utilities/logging")
+
+# Add demo
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/demos)
+
+# Add MCU flags, set using board cmake
+set(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} ${MCU_C_FLAGS})
+set(CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS} ${MCU_LINKER_FLAGS})
+
+# Set middleware specific path's
+set(FREERTOS_DIRECTORY ${FREERTOS_PATH} CACHE STRING "FreeRTOS Path")
+set(CONFIG_DIRECTORY ${BOARD_DEMO_CONFIG_PATH} CACHE STRING "Demo config")
+set(FREERTOS_PORT_DIRECTORY ${BOARD_DEMO_FREERTOS_PORT_PATH} CACHE STRING "FreeRTOS port")
+
+# Add middleware
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/libs/azure-iot-middleware-freertos)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,8 +14,8 @@ include(common/utilities)
 
 # Set FreeRTOS path
 if(NOT FREERTOS_PATH)
-    free_rtos_fetch()
     message(STATUS "No FREERTOS_PATH specified: Fetching FreeRTOS to path <${FREERTOS_PATH}>")
+    free_rtos_fetch()
 endif()
 
 find_package(FreeRTOS REQUIRED)
--- a/cmake/FindFreeRTOS.cmake
+++ b/cmake/FindFreeRTOS.cmake
@@ -1,0 +1,118 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+# Modified version from https://github.com/ObKo/stm32-cmake
+
+if(NOT FreeRTOS_FIND_COMPONENTS)
+    set(FreeRTOS_FIND_COMPONENTS
+        ARM_CM0 ARM_CM3 ARM_CM4F ARM_CM7 Posix
+    )
+endif()
+list(REMOVE_DUPLICATES FreeRTOS_FIND_COMPONENTS)
+
+set(FreeRTOS_HEAPS 1 2 3 4 5)
+
+if(NOT FREERTOS_PATH)
+    set(FREERTOS_PATH /opt/FreeRTOS CACHE PATH "Path to FreeRTOS")
+    message(STATUS "No FREERTOS_PATH specified using default: ${FREERTOS_PATH}")
+endif()
+
+find_path(FreeRTOS_COMMON_INCLUDE
+    NAMES FreeRTOS.h
+    PATHS "${FREERTOS_PATH}" "${FREERTOS_PATH}/FreeRTOS" 
+    PATH_SUFFIXES  "Source/include"
+    NO_DEFAULT_PATH
+)
+list(APPEND FreeRTOS_INCLUDE_DIRS "${FreeRTOS_COMMON_INCLUDE}")
+
+find_path(FreeRTOS_SOURCE_DIR
+    NAMES tasks.c
+    PATHS "${FREERTOS_PATH}" "${FREERTOS_PATH}/FreeRTOS" 
+    PATH_SUFFIXES  "Source"
+    NO_DEFAULT_PATH
+)
+if(NOT (TARGET FreeRTOS))
+    add_library(FreeRTOS INTERFACE IMPORTED)
+    target_sources(FreeRTOS INTERFACE 
+        "${FreeRTOS_SOURCE_DIR}/tasks.c"
+        "${FreeRTOS_SOURCE_DIR}/list.c"
+        "${FreeRTOS_SOURCE_DIR}/queue.c"
+    )
+    target_include_directories(FreeRTOS INTERFACE "${FreeRTOS_COMMON_INCLUDE}")
+endif()
+
+if(NOT (TARGET FreeRTOS::Coroutine))
+    add_library(FreeRTOS::Coroutine INTERFACE IMPORTED)
+    target_sources(FreeRTOS::Coroutine INTERFACE "${FreeRTOS_SOURCE_DIR}/croutine.c")
+    target_link_libraries(FreeRTOS::Coroutine INTERFACE FreeRTOS)
+endif()
+
+if(NOT (TARGET FreeRTOS::EventGroups))
+    add_library(FreeRTOS::EventGroups INTERFACE IMPORTED)
+    target_sources(FreeRTOS::EventGroups INTERFACE "${FreeRTOS_SOURCE_DIR}/event_groups.c")
+    target_link_libraries(FreeRTOS::EventGroups INTERFACE FreeRTOS)
+endif()
+
+if(NOT (TARGET FreeRTOS::StreamBuffer))
+    add_library(FreeRTOS::StreamBuffer INTERFACE IMPORTED)
+    target_sources(FreeRTOS::StreamBuffer INTERFACE "${FreeRTOS_SOURCE_DIR}/stream_buffer.c")
+    target_link_libraries(FreeRTOS::StreamBuffer INTERFACE FreeRTOS)
+endif()
+
+if(NOT (TARGET FreeRTOS::Timers))
+    add_library(FreeRTOS::Timers INTERFACE IMPORTED)
+    target_sources(FreeRTOS::Timers INTERFACE "${FreeRTOS_SOURCE_DIR}/timers.c")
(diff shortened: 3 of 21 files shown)


cmake/FindFreeRTOS.cmake
```diff
@@ -0,0 +1,118 @@
...
+    NO_DEFAULT_PATH
+)
+if(NOT (TARGET FreeRTOS))
+    add_library(FreeRTOS INTERFACE IMPORTED)
+    target_sources(FreeRTOS INTERFACE 
+        "${FreeRTOS_SOURCE_DIR}/tasks.c"
+        "${FreeRTOS_SOURCE_DIR}/list.c"
+        "${FreeRTOS_SOURCE_DIR}/queue.c"
+    )
+    target_include_directories(FreeRTOS INTERFACE "${FreeRTOS_COMMON_INCLUDE}")
```
danewalton, line 41: nit: maybe make this `FreeRTOS::Kernel` or `FreeRTOS::FreeRTOS`

cmake/FindFreeRTOSPlus.cmake
```diff
@@ -0,0 +1,149 @@
...
+        ${FreeRTOSPlus_PATH}/ThirdParty/mbedtls/include)
+    target_include_directories(FreeRTOSPlus::ThirdParty::mbedtls INTERFACE
+        ${FreeRTOSPlus_PATH}/Source/Utilities/mbedtls_freertos)
+    target_compile_definitions(FreeRTOSPlus::ThirdParty::mbedtls INTERFACE
+        MBEDTLS_CONFIG_FILE="mbedtls_config.h")
+endif()
+
+if(NOT (TARGET FreeRTOSPlus::TCPIP))
+    add_library(FreeRTOSPlus::TCPIP INTERFACE IMPORTED)
+    target_sources(FreeRTOSPlus::TCPIP INTERFACE
```
danewalton, line 125: does everything here need to be INTERFACE? INTERFACE is only public facing (not PRIVATE). If you make it PUBLIC it's in INTERFACE and PRIVATE

demos/projects/PC/linux/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+include(${CMAKE_CURRENT_SOURCE_DIR}/gcc_flags.cmake)
+
+# include config path as global
+include_directories(${BOARD_DEMO_CONFIG_PATH})
+
+# Add port specific source file
+target_sources(FreeRTOSPlus::TCPIP::PORT INTERFACE 
+    ${FreeRTOSPlus_PATH}/Source/FreeRTOS-Plus-TCP/portable/BufferManagement/BufferAllocation_2.c
+    ${FreeRTOSPlus_PATH}/Source/FreeRTOS-Plus-TCP/portable/NetworkInterface/linux/NetworkInterface.c
+    ../../../common/transport/sockets_wrapper_freertos_tcpip.c)
```
danewalton, line 18: is there a variable this can be relative from? like `${FreeRTOSPlus_PATH}`

demos/projects/PC/linux/config/FreeRTOSIPConfig.h
```diff
@@ -0,0 +1,308 @@
...
+#define ipconfigTCP_HANG_PROTECTION         ( 1 )
+#define ipconfigTCP_HANG_PROTECTION_TIME    ( 30 )
+
+/* Include support for TCP keep-alive messages. */
+#define ipconfigTCP_KEEP_ALIVE              ( 1 )
+#define ipconfigTCP_KEEP_ALIVE_INTERVAL     ( 20 ) /* in seconds */
+
+#define portINLINE                          __inline
+
+#endif /* FREERTOS_IP_CONFIG_H */
```
danewalton, line 308: newline

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+ * 2. Define the LIBRARY_LOG_NAME and LIBRARY_LOG_LEVEL macros depending on
+ * the logging configuration for DEMO.
+ * 3. Include the header file "logging_stack.h", if logging is enabled for DEMO.
+ */
+
+#include "logging_levels.h"
+
+/* Logging configuration for the Demo. */
+#ifndef LIBRARY_LOG_NAME
+    #define LIBRARY_LOG_NAME    "MQTTDemo"
```
danewalton, line 25: ?
```suggestion
    #define LIBRARY_LOG_NAME    "AzureIoTDemo"
```

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+
+#ifdef democonfigENABLE_DPS_SAMPLE
+
+/**
+ * @brief Provisioning service endpoint.
+ *
+ * @note https://docs.microsoft.com/en-us/azure/iot-dps/concepts-service#service-operations-endpoint
+ * 
+ */
+#define democonfigENDPOINT                  "<YOUR DPS ENDPOINT HERE>"
```
danewalton, line 65: I've been thinking should we just have the global provisioning url here? It's highly unlikely this will change and it adds another step

demos/projects/PC/linux/main.c
```diff
@@ -0,0 +1,462 @@
...
+    prvSRand( ( uint32_t ) xTimeNow );
+    LogDebug( ( "Random numbers: %08X %08X %08X %08X\n", ipconfigRAND32(), ipconfigRAND32(), ipconfigRAND32(), ipconfigRAND32() ) );
+}
+/*-----------------------------------------------------------*/
+
+#if ( ipconfigUSE_LLMNR != 0 ) || ( ipconfigUSE_NBNS != 0 ) || ( ipconfigDHCP_REGISTER_HOSTNAME == 1 )
+
+    const char * pcApplicationHostnameHook( void )
+    {
+        /* Assign the name "FreeRTOS" to this network node.  This function will
```
danewalton, line 270: The "FreeRTOS" part is now wrong. Just delete the first sentence?

danewalton (approved):

demos/projects/PC/linux/config/demo_config.h
hihigupt: (y)

cmake/FindFreeRTOS.cmake
hihigupt: for now I want to keep it in sync with https://github.com/ObKo/stm32-cmake version

demos/projects/PC/linux/CMakeLists.txt
hihigupt: removed

cmake/FindFreeRTOSPlus.cmake
hihigupt: discuss offline... for now I will keep it same

cmake/PC_gcc.cmake
```diff
@@ -0,0 +1,14 @@
...
+
+get_filename_component(CMAKE_TOOL_DIR ${CMAKE_CURRENT_LIST_FILE} DIRECTORY)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_TOOL_DIR})
+
+set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+find_program(CMAKE_C_COMPILER NAMES gcc)
+find_program(CMAKE_CXX_COMPILER NAMES g++)
+find_program(CMAKE_ASM_COMPILER NAMES gcc)
+
+set(CMAKE_EXECUTABLE_SUFFIX_C   .elf)
```
danewalton, line 12: should these still be elf?

hihigupt: File format is elf produce by gcc in default. 

Command:
[bash]$ file build/demos/projects/PC/linux/iot-middleware-sample.elf

build/demos/projects/PC/linux/iot-middleware-sample.elf: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=9e39e3316418d874b2c5b041bb2afcb9db179348, with debug_info, not stripped
h

.github/scripts/ci_tests.sh
```diff
@@ -0,0 +1,46 @@
+#! /bin/bash
```
jspaith, line 1: Nit - is .github the directory stuff something like this should live?  I've normally seen this for bread&butter hub setup and the tests live "someplace else."  If this is a convention I'm not aware of it's fine as is, but want to make sure we're in alignment.

.github/scripts/ci_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+set -o errexit # Exit if command failed.
+set -o nounset # Exit if variable not set.
+set -o pipefail # Exit if pipe failed.
+
+echo -e "::group::FreeRTOS Source"
+
+if [ ! -d "libs/FreeRTOS" ]; then
+    git clone https://github.com/FreeRTOS/FreeRTOS.git libs/FreeRTOS
```
jspaith, line 13: If / when something goes wrong on a gate, how are we going to debug?  If every line - specifically like `git clone blah` - is automatically echoed then perfect before its executed, that's the best.  Otherwise we should consider adding a few `echo beginning git clone` | `starting cmake` on longer running commands that may hang or fail without flushing stdout/stderr.

.github/scripts/ci_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+    pushd libs/FreeRTOS
+    git checkout -b c8fa483b68c6c1149c2a7a8bc1e901b38860ec9b
```
jspaith, line 15: Nit1 - consider making this an environment variable at top of file, so it's one-stop shopping when we need to change.


Nit2 - is there a way we could make this "more global" so that every .md/bash/etc. from everywhere referenced the one truth state of this?  I worry this is the type of thing that gets out of sync, where half the files are using 1 commit ID and others are using a different over time. 

Or it breaks laws of physics or becomes even worse to try and "one-truth-source" we should make sure we have clear process for updating this, and as few references of these as we can still do cleanly.

.github/scripts/ci_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+    git submodule sync
+    git submodule update --init --recursive --depth=1
+    popd
+fi
+TEST_FREERTOS_SRC=`pwd`/libs/FreeRTOS
+
+function binary_exists()
```
jspaith, line 22: Nit - recommend this be something like "exit_if_binary_does_not_exist()" or something better word-smithed so the side effect of script ending is a bit more clear.  Though maybe in bash land this is common practice??

.github/scripts/ci_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+    local vendor=$1
+    local board=$2
+    local outdir=$3
+
+    cmake -G Ninja -DBOARD=$board -DVENDOR=$vendor -B$outdir -DFREERTOS_PATH=$TEST_FREERTOS_SRC .
+    cmake --build $outdir
+}
+
+echo -e "::group::Building sample for linux port"
+build "PC" "linux" "build_pc_linux"
```
jspaith, line 45: What is `build` here - not clear what function this is or built-in linux command this would invoke?

CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+
+# include common utilities
+include(common/utilities)
+
+# Set FreeRTOS path
+if(NOT FREERTOS_PATH)
+    free_rtos_fetch()
+    message(STATUS "No FREERTOS_PATH specified: Fetching FreeRTOS to path <${FREERTOS_PATH}>")
```
jspaith, line 18: Nit - shouldn't the message() come before the free_rtos_fetch command?

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+/* Logging configuration for the Demo. */
+#ifndef LIBRARY_LOG_NAME
+    #define LIBRARY_LOG_NAME    "AzureIoTDemo"
+#endif
+
+#ifndef LIBRARY_LOG_LEVEL
+    #define LIBRARY_LOG_LEVEL    LOG_INFO
+#endif
+
+/* Prototype for the function used to print to console on Windows simulator
```
jspaith, line 32: References Windows simulator but this is under the Linux folder.

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+#endif // democonfigENABLE_DPS_SAMPLE
+
+/**
+ * @brief IoTHub device Id.
+ *
+ */
+#define democonfigDEVICE_ID                 "<YOUR DEVICE ID HERE>"
+
+/**
+ * @brief IoTHub module Id.
```
jspaith, line 93: We should call out that this is optional, correct?  And anyplace else with similar header?

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+ * @brief IoTHub hostname.
+ *
+ */
+#define democonfigHOSTNAME                  "<YOUR IOT HUB HOSTNAME HERE>"
+
+/**
+ * @brief Device symmetric key
+ *
+ */
+#define democonfigDEVICE_SYMMETRIC_KEY      "<Symmetric key>"
```
jspaith, line 107: Could we make it more clear - eventually, OK if it's not this PR but we should track work somehow - that it's either Sym key or else x509?  That doesn't necessarly jump out at me way that the code is currently commented.

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+"-----END CERTIFICATE-----\r\n"
+
+/**
+ * @brief An option to disable Server Name Indication.
+ *
+ * @note When using a local Mosquitto server setup, SNI needs to be disabled
+ * for an MQTT broker that only has an IP address but no hostname. However,
+ * SNI should be enabled whenever possible.
+ */
+#define democonfigDISABLE_SNI              ( pdFALSE )
```
jspaith, line 154: Ugh - in perfect world we wouldn't have this type of flag that user flips on to get things going and then forgets it and ships.

Could we put some sort of thing in main() where like

```c
#if (democonfigDISABLE_SNI == pdTrue)
   LogWarn("You've disabled SNI for this test.  Some sort of properly word-smithed callout on this");  
#endif
```

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * @brief The name of the operating system that the application is running on.
+ * The current value is given as an example. Please update for your specific
+ * operating system.
+ */
+#define democonfigOS_NAME                   "FreeRTOS"
```
jspaith, line 161: Do we need this as an option, or at least the part encouraged developer to update for specific OS?

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * @brief The version of the operating system that the application is running
+ * on. The current value is given as an example. Please update for your specific
+ * operating system version.
+ */
+#define democonfigOS_VERSION                tskKERNEL_VERSION_NUMBER
```
jspaith, line 168: same question as above RE whether we should encoruage app developer to be changing this given our FReeRTOS centric approach?  We've already set them up with right thing for freertos, correct?

jspaith (approved):

demos/projects/PC/linux/config/demo_config.h
hihigupt: removed

demos/projects/PC/linux/config/demo_config.h
hihigupt: not required... this is some of stuff came from FreeRTOS classic demo example

demos/projects/PC/linux/config/demo_config.h
hihigupt: SNI should not be disable, I will remove this option and make it always true in sample

demos/projects/PC/linux/config/demo_config.h
hihigupt: sample does have check that either Symmetric key or Cert is define.

https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/dae616d0c665684590f60252ad6d6442608d4531/demos/sample_azure_iot/sample_azure_iot.c#L46

demos/projects/PC/linux/config/demo_config.h
hihigupt: Updated

.github/scripts/ci_tests.sh
hihigupt: build is function define in this file : 35. I have updated function name to be sample_build, just to make it more obvious

.github/scripts/ci_tests.sh
hihigupt: all the github Actions get defined in this folder (.github/workflow/). So all the scripts that are getting used in this workflow are kept under it only.

.github/scripts/ci_tests.sh
hihigupt: git clone does the echo for you:

https://github.com/Azure-Samples/iot-middleware-freertos-samples/runs/2760640942?check_suite_focus=true#step:5:9

.github/scripts/ci_tests.sh
hihigupt: Ack. I created global variable for TEST FreeRTOS commit id.

Right now there are two places:
1. Test FreeRTOS
2. Cmake FreeRTOS commit and it can be override by caller (Test is caller of cmake).

Not sure of any better way to do it.

demos/projects/PC/linux/config/demo_config.h
hihigupt: Removed

.github/scripts/ci_tests.sh
jspaith: I only saw the 2 places too - provided that's it (and i'm hoping it will be) that should be manegaable.  At some point on a future PR you might consider a comment around the commitID that "If you update this, update all of them" just as a reminder to potential future mainteners.  Consider only as i said...

PR #26, opened 2021-06-16
Change init_linux_port_vm_network.sh to make it safe to use

ewertons:
## Purpose
Make sure init_linux_port_vm_network.sh does not destroy information in the host it runs on.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/.github/scripts/init_linux_port_vm_network.sh
+++ b/.github/scripts/init_linux_port_vm_network.sh
@@ -3,12 +3,15 @@
 # Copyright (c) Microsoft. All rights reserved.
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #
-# This script sets up virtual interface (veth1 and veth0 ) with DHCP configure that can be used for linux port.
+# This script sets up virtual interface (rtosveth1 and rtosveth0) with DHCP configure that can be used for linux port.
+
+set -o errexit  # Exit if command failed.
+set -o nounset  # Exit if variable not set.
+set -o pipefail # Exit if pipe failed.
 
 getNextAvailableIpRange() {
     for (( i=1; i<=255; i++ )); do
         match=`ifconfig -a | grep -o -E "inet $1\.$2\.$i\.[0-9]+"`
-        #echo "$match"
 
         if [[ "$match" == "" ]]; then
             echo $i
@@ -21,9 +24,9 @@ if [[ "$1" == "--help" || "$1" == "-h" ]]; then
     echo "Creates a virtual network interface for the freertos simulator."
     echo "Use option --clean to undo any changes."
 elif [[ "$1" == "--clean" ]]; then
-    ifconfig veth0 down
-    ifconfig veth1 down
-    ip link delete veth0
+    ifconfig rtosveth0 down
+    ifconfig rtosveth1 down
+    ip link delete rtosveth0
     rm -f /etc/dhcp/dhcpd.conf
     rm -f /etc/default/isc-dhcp-server
 
@@ -49,11 +52,11 @@ else
 
     ip_octet=$( getNextAvailableIpRange 192 168 )
 
-    ip link add veth0 type veth peer name veth1
-    ifconfig veth0 192.168.$ip_octet.1 netmask 255.255.255.0 up
-    ifconfig veth1 up
-    ethtool --offload veth0 tx off
-    ethtool --offload veth1 tx off
+    ip link add rtosveth0 type veth peer name rtosveth1
+    ifconfig rtosveth0 192.168.$ip_octet.1 netmask 255.255.255.0 up
+    ifconfig rtosveth1 up
+    ethtool --offload rtosveth0 tx off
+    ethtool --offload rtosveth1 tx off
 
     sh -c "cat > /etc/dhcp/dhcpd.conf" <<EOT
     subnet 192.168.$ip_octet.0 netmask 255.255.255.0 {
@@ -67,7 +70,7 @@ else
 EOT
 
     sh -c "cat > /etc/default/isc-dhcp-server" <<EOT
-    INTERFACES="veth0"
+    INTERFACES="rtosveth0"
 EOT
     service isc-dhcp-server restart
 
@@ -77,7 +80,7 @@ EOT
     iptables -F
     iptables -t nat -F
     iptables -t nat -A POSTROUTING -s 192.168.$ip_octet.0/24 -o eth0 -j MASQUERADE
-    iptables -A FORWARD -d 192.168.$ip_octet.0/24 -o veth0 -j ACCEPT
+    iptables -A FORWARD -d 192.168.$ip_octet.0/24 -o rtosveth0 -j ACCEPT
     iptables -A FORWARD -s 192.168.$ip_octet.0/24 -j ACCEPT
 
     #Debug
--- a/.github/scripts/init_linux_port_vm_network.sh
+++ b/.github/scripts/init_linux_port_vm_network.sh
@@ -5,37 +5,85 @@
 #
 # This script sets up virtual interface (veth1 and veth0 ) with DHCP configure that can be used for linux port.
 
-ip link add veth0 type veth peer name veth1
-ifconfig veth0 192.168.1.1 netmask 255.255.255.0 up
-ifconfig veth1 up
-ethtool --offload veth0 tx off
-ethtool --offload veth1 tx off
+getNextAvailableIpRange() {
+    for (( i=1; i<=255; i++ )); do
+        match=`ifconfig -a | grep -o -E "inet $1\.$2\.$i\.[0-9]+"`
+        #echo "$match"
 
-sh -c "cat > /etc/dhcp/dhcpd.conf" <<EOT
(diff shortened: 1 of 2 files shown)


.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -5,37 +5,85 @@
...
+else
+    if [ -f "/etc/dhcp/dhcpd.conf" ]; then
+        cp -v /etc/dhcp/dhcpd.conf /etc/dhcp/dhcpd.conf.bkp
+    fi
+
+    if [ -f "/etc/default/isc-dhcp-server" ]; then
+        cp -v "/etc/default/isc-dhcp-server" "/etc/default/isc-dhcp-server.bkp"
+    fi
+
+    ip_octet=$( getNextAvailableIpRange 192 168 )
```
danewalton, line 50: curious what this does. does this provide any extra safety compared to what the script was before

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -5,37 +5,85 @@
...
-    range 192.168.1.100 192.168.1.200;
-    option routers 192.168.1.1;
-    option domain-name-servers 1.1.1.1;
-    option broadcast-address 192.168.1.255;
-    default-lease-time 600;
-    max-lease-time 7200;
+getNextAvailableIpRange() {
+    for (( i=1; i<=255; i++ )); do
+        match=`ifconfig -a | grep -o -E "inet $1\.$2\.$i\.[0-9]+"`
+        #echo "$match"
```
danewalton, line 11: remove?

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -5,37 +5,85 @@
 #
```
hihigupt, line 5: should we add 
set -o errexit # Exit if command failed. 
set -o nounset # Exit if variable not set. 
set -o pipefail # Exit if pipe failed.


https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/06d8de660ca1a90e8ca8c2b03eaf4161060decfc/.github/scripts/ci_tests.sh#L6

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -5,37 +5,85 @@
...
+            break
+        fi
+    done
 }
+
+if [[ "$1" == "--help" || "$1" == "-h" ]]; then
+    echo "Creates a virtual network interface for the freertos simulator."
+    echo "Use option --clean to undo any changes."
+elif [[ "$1" == "--clean" ]]; then
+    ifconfig veth0 down
```
hihigupt, line 24: how about letting user pass interface name ? and by default we name the interface as veth0 if not passed?

how to set default value.
https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/06d8de660ca1a90e8ca8c2b03eaf4161060decfc/.github/scripts/fetch_freertos.sh#L12****

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -5,37 +5,85 @@
...
+    ifconfig veth1 down
+    ip link delete veth0
```
hihigupt, line 26: do we need to delete ? I do not see this in older script.

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -5,37 +5,85 @@
...
+        mv "/etc/dhcp/dhcpd.conf.bkp" "/etc/dhcp/dhcpd.conf"
+    fi
+
+    if [ -f "/etc/default/isc-dhcp-server.bkp" ]; then
+        mv "/etc/default/isc-dhcp-server.bkp" "/etc/default/isc-dhcp-server"
+    fi
+
+    service isc-dhcp-server restart
+
+    iptables-restore < /opt/iptables.backup
```
hihigupt, line 40: is it ok to assume that this file exists /opt/iptables.backup?

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -5,37 +5,85 @@
...
+    iptables -A FORWARD -d 192.168.$ip_octet.0/24 -o veth0 -j ACCEPT
+    iptables -A FORWARD -s 192.168.$ip_octet.0/24 -j ACCEPT
+
+    #Debug
+    ip link
+    ip addr
+    iptables -t nat -L -n
+    cat /etc/dhcp/dhcpd.conf
+    cat /etc/default/isc-dhcp-server
+fi
```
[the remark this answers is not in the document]
ewertons: Done. Thaks.

.github/scripts/init_linux_port_vm_network.sh
ewertons: Yes, the previous script would just use 192.168.1.x for the new virtual eth interface, possibly conflicting with an existing interface (in case for example if you have 192.168.1.x set in your home lan - _like me_).
This searches for the next ip range (changing the 3rd octect) available (ie, not used by the local network interfaces).

.github/scripts/init_linux_port_vm_network.sh
ewertons: Done.

.github/scripts/init_linux_port_vm_network.sh
ewertons: If the customer is running the --clean, it is assumed they have run the bare script before. If the file is not there it's safe because the command will fail. I'll double check that to be sure.

.github/scripts/init_linux_port_vm_network.sh
ewertons: Sounds like a good idea.

.github/scripts/init_linux_port_vm_network.sh
ewertons: We could... I dont' think it's a high priority, because it's unlikely the system will have a veth0 and veth1 already. I'll sync with you offline.

.github/scripts/init_linux_port_vm_network.sh
ewertons: Yeah, it's safe. The command fails (and the current iptables config does not get affected) if the file is not present.

```bash
ewertons@ewertons-lcx1g3:~/code/work/s1/iot-middleware-freertos-samples-ew$ sudo iptables-restore < ./test.bkp 
bash: ./test.bkp: No such file or directory 
```

.github/scripts/init_linux_port_vm_network.sh
ewertons: Talked to Himanshu, we will not add option to set name of virtual interface, but we will rename the default names from veth0 to rtosveth0 and rtosveth1.

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -3,39 +3,90 @@
...
+        match=`ifconfig -a | grep -o -E "inet $1\.$2\.$i\.[0-9]+"`
+
+        if [[ "$match" == "" ]]; then
+            echo $i
+            break
+        fi
+    done
 }
+
+if [[ "$1" == "--help" || "$1" == "-h" ]]; then
```
hihigupt, line 23: It looks like if you run this script after adding 8-10 lines with no arguments, script should fail with ": $1: unbound variable". Can you please confirm.

.github/scripts/init_linux_port_vm_network.sh
hihigupt: This will cause script to finish in unclean maner. Can we check if file exists.

.github/scripts/init_linux_port_vm_network.sh
```diff
@@ -3,39 +3,90 @@
...
-cat /etc/dhcp/dhcpd.conf
-cat /etc/default/isc-dhcp-server
+    service isc-dhcp-server restart
+
+    iptables-save > /opt/iptables.backup
+
+    sysctl net.ipv4.ip_forward=1
+    iptables -F
+    iptables -t nat -F
+    iptables -t nat -A POSTROUTING -s 192.168.$ip_octet.0/24 -o eth0 -j MASQUERADE
```
hihigupt, line 82: :( we are assuming internet ethernet interface is eth0, we can't cover all the scenario. Can we please document this at the top of file.

.github/scripts/init_linux_port_vm_network.sh
ewertons: Fixed.

.github/scripts/init_linux_port_vm_network.sh
ewertons: Good catch. I added a function to find which interface has connectivity.

hihigupt (approved):

danewalton (approved):

PR #28, opened 2021-06-17
pnp sample

danewalton:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,9 @@
+{
+    "image": "tizho/azurertos-regression",
+	"runArgs": [ "--cap-add=NET_ADMIN"],
+	"postStartCommand": "/init_vm_network.sh",
```
hihigupt, line 4: this script name is changed

demos/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ string(TOUPPER ${BOARD} BOARD_U)
 if(NOT (TARGET SAMPLE::AZUREIOT))
     add_library(SAMPLE::AZUREIOT INTERFACE IMPORTED)
     target_sources(SAMPLE::AZUREIOT INTERFACE 
-        ${CMAKE_CURRENT_SOURCE_DIR}/sample_azure_iot/sample_azure_iot.c)
+        ${CMAKE_CURRENT_SOURCE_DIR}/sample_azure_iot_pnp/sample_azure_iot_pnp.c)
```
hihigupt, line 15: how about we create one more target of PNP sample and let project decide, which one to pick based on some CMAKE variable pass?

demos/projects/PC/linux/main.c
```diff
@@ -241,6 +241,17 @@ static void prvSRand( UBaseType_t ulSeed )
 }
 /*-----------------------------------------------------------*/
 
+UBaseType_t uxRand( void )
```
hihigupt, line 244: >uxRand [](http://example.com/codeflow?start=12&length=6)

weird, is this required?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+/* Standard includes. */
+#include <string.h>
+#include <stdio.h>
+#include <time.h>
```
hihigupt, line 7: this looks to be wrong dependency, as we depend upon getunittime api to get time, and that api is provided by sample project ports .... if true for other sample, can we remove from there also.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "task.h"
+
+/* Demo Specific configs. */
+#include "demo_config.h"
```
hihigupt, line 14: can we move this to line 32?

demos/projects/PC/linux/main.c
danewalton: Yea if you enable `LOG_DEBUG` instead of just `LOG_INFO` it needs this function

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+#define sampleazureiotTELEMETRY_NAME                      "temperature"
+
+/**
+ * @brief Property Values
+ */
+#define sampleazureiotPROPERTY_STATUS_SUCCESS             200
+#define sampleazureiotPROPERTY_SUCCESS                    "success"
+#define sampleazureiotPROPERTY_TARGET_TEMPERATURE_TEXT    "targetTemperature"
+#define sampleazureiotPROPERTY_MAX_TEMPERATURE_TEXT       "maxTempSinceLastReboot"
+static uint8_t ucPropertyPayloadBuffer[ 256 ];
```
hihigupt, line 111: how about structuring the file with following format:

#include


#define

export

static

static fn()

fn()

?

danewalton: Here I think it's nice bunching the values together that are related to a specific feature. At a glance, you can see all the things being used for a specific feature. If there is significant passion for grouping by var type I could be swayed but I think this keeps things nicely organized

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: needed for this guy here
https://github.com/Azure-Samples/iot-middleware-freertos-samples/pull/28/files/708592629b6b5f3bc9f2faff509c6f44161ec066#diff-b44ce38db1c6040a3fd014270977865367a1222b4ae0c415c6c5fcdab47de58aR301

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+ * @brief Gets the IoT Hub endpoint and deviceId from Provisioning service.
+ *   This function will block for Provisioning service for result or return failure.
+ *
+ * @param[in] pXNetworkCredentials  Network credential used to connect to Provisioning service
+ * @param[out] ppucIothubHostname  Pointer to uint8_t* IoT Hub hostname return from Provisioning Service
+ * @param[in,out] pulIothubHostnameLength  Length of hostname
+ * @param[out] ppucIothubDeviceId  Pointer to uint8_t* deviceId return from Provisioning Service
+ * @param[in,out] pulIothubDeviceIdLength  Length of deviceId
+ */
+    static uint32_t prvIoTHubInfoGet( NetworkCredentials_t * pXNetworkCredentials,
```
hihigupt, line 225: >prvIoTHubInfoGet [](http://example.com/codeflow?start=20&length=16)

how about moving these common function between samples in common folder? and for all other generic functions?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+ */
+static uint8_t ucMQTTMessageBuffer[ democonfigNETWORK_BUFFER_SIZE ];
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Generate max min payload.
+ */
+int32_t prvInvokeMaxMinCommand( const uint8_t * pucPayload,
+                                uint32_t ulPayloadLength )
```
hihigupt, line 271: can this be static function?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+
+    xResult = AzureIoTJSONReader_GetTokenString( &xReader,
+                                                 ucCommandStartTimeValueBuffer,
+                                                 sizeof( ucCommandStartTimeValueBuffer ),
+                                                 &ulSinceTimeLength );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+
+    /* Get the current time as a string. */
+    time( &xRawTime );
```
hihigupt, line 301: >time [](http://example.com/codeflow?start=4&length=4)

is it possible to use getunixtime?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+    time_t xRawTime;
+    struct tm * pxTimeInfo;
+    size_t xEndTimeLength;
+
+    /* Get the start time */
+    xResult = AzureIoTJSONReader_Init( &xReader, pucPayload, ulPayloadLength );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTJSONReader_NextToken( &xReader );
+    configASSERT( xResult == eAzureIoTSuccess );
```
hihigupt, line 285: should this function return error? just to report 501 error for method response.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+    xEndTimeLength = strftime(
+        ucCommandEndTimeValueBuffer,
+        sizeof( ucCommandEndTimeValueBuffer ),
+        sampleazureiotDATE_TIME_FORMAT,
+        pxTimeInfo );
+
+    AzureIoTJSONWriter_t xWriter;
+
+    xResult = AzureIoTJSONWriter_Init( &xWriter, ucCommandPayloadBuffer, sizeof( ucCommandPayloadBuffer ) );
+    configASSERT( xResult == eAzureIoTSuccess );
```
hihigupt, line 312: its little hard to follow, how this function is communicating result back to caller.  Side effect of this function is some global memory, whose bytes used you are returning.  Better way to express will be to have this function take the reader and output writer.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+    /* Get the current time as a string. */
+    time( &xRawTime );
+    pxTimeInfo = localtime( &xRawTime );
+    xEndTimeLength = strftime(
+        ucCommandEndTimeValueBuffer,
+        sizeof( ucCommandEndTimeValueBuffer ),
+        sampleazureiotDATE_TIME_FORMAT,
+        pxTimeInfo );
+
+    AzureIoTJSONWriter_t xWriter;
```
hihigupt, line 309: >AzureIoTJSONWriter_t [](http://example.com/codeflow?start=4&length=20)

lets keep the formating similar to other places in this repo, to define all the variable upfront in function entry.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+
+    xResult = AzureIoTJSONWriter_Init( &xWriter, ucCommandPayloadBuffer, sizeof( ucCommandPayloadBuffer ) );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTJSONWriter_AppendBeginObject( &xWriter );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTJSONWriter_AppendPropertyWithDoubleValue( &xWriter, sampleazureiotCOMMAND_MAX_TEMP, strlen( sampleazureiotCOMMAND_MAX_TEMP ), xDeviceMaximumTemperature, sampleazureiotDOUBLE_DECIMAL_PLACE_DIGITS );
```
hihigupt, line 317: nit: break the line

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    return AzureIoTJSONWriter_GetBytesUsed( &xWriter );
+}
+
+/**
+ * @brief Cloud message callback handler
+ */
+static void prvHandleCloudMessage( AzureIoTHubClientCloudToDeviceMessageRequest_t * pxMessage,
+                                   void * pvContext )
```
hihigupt, line 342: remove this ... as PnP does not have this concept.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+    LogInfo( ( "Command payload : %.*s \r\n",
+               pxMessage->ulPayloadLength,
+               pxMessage->pvMessagePayload ) );
+
+    int32_t ulCommandPayloadLength;
+
+    AzureIoTHubClient_t * xHandle = ( AzureIoTHubClient_t * ) pvContext;
+
+    if( strncmp( sampleazureiotCOMMAND_MAX_MIN_REPORT, pxMessage->pucCommandName, strlen( sampleazureiotCOMMAND_MAX_MIN_REPORT ) ) == 0 )
+    {
```
hihigupt, line 367: first compare the length are equal and then compare the memory

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+            LogInfo( ( "Error generating command payload" ) );
+        }
+    }
+    else
+    {
+        /* Not for max min report (not for this device) */
+        LogInfo( ( "Received command is not for this device" ) );
+
+        if( AzureIoTHubClient_SendCommandResponse( xHandle, pxMessage, 404,
+                                                   sampleazureiotCOMMAND_EMPTY_PAYLOAD,
```
hihigupt, line 391: >sampleazureiotCOMMAND_EMPTY_PAYLOAD [](http://example.com/codeflow?start=51&length=35)

we can pass NULL

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+        /* Is for max min report */
+        ulCommandPayloadLength = prvInvokeMaxMinCommand( pxMessage->pvMessagePayload, pxMessage->ulPayloadLength );
+
+        if( ulCommandPayloadLength > 0 )
+        {
+            if( AzureIoTHubClient_SendCommandResponse( xHandle, pxMessage, 200,
+                                                       ucCommandPayloadBuffer,
+                                                       ulCommandPayloadLength ) != eAzureIoTSuccess )
+            {
+                LogInfo( ( "Error sending command response" ) );
```
hihigupt, line 377: these are error level loggings

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+            }
+        }
+        else
+        {
+            LogInfo( ( "Error generating command payload" ) );
```
hihigupt, line 382: >LogInfo [](http://example.com/codeflow?start=12&length=7)

warning.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
hihigupt: please check all the places, for the level of log

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+ * @brief Direct command message callback handler
+ */
+static void prvHandleCommand( AzureIoTHubClientCommandRequest_t * pxMessage,
+                              void * pvContext )
+{
+    LogInfo( ( "Command payload : %.*s \r\n",
+               pxMessage->ulPayloadLength,
+               pxMessage->pvMessagePayload ) );
+
+    int32_t ulCommandPayloadLength;
```
hihigupt, line 362: >i [](http://example.com/codeflow?start=4&length=1)

nit : first line

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+        /* Reset JSON reader to the beginning */
+        xResult = AzureIoTJSONReader_Init( &xReader, pxMessage->pvMessagePayload, pxMessage->ulPayloadLength );
+        configASSERT( xResult == eAzureIoTSuccess );
+
+        while( ( xResult = AzureIoTHubClientProperties_GetNextComponentProperty( &xAzureIoTHubClient, &xReader,
+                                                                                 pxMessage->xMessageType, xPropertyType,
+                                                                                 &pucComponentName, &ulComponentNameLength ) ) == eAzureIoTSuccess )
+        {
+            xResult = AzureIoTJSONReader_TokenType( &xReader, &xTokenType );
+            configASSERT( xResult == eAzureIoTSuccess );
```
hihigupt, line 433: this is agressing way of handling these error ? Parsing error should be reported back and possibly back to cloud with some error code.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+                                                   sampleazureiotCOMMAND_EMPTY_PAYLOAD,
+                                                   strlen( sampleazureiotCOMMAND_EMPTY_PAYLOAD ) ) != eAzureIoTSuccess )
+        {
+            LogInfo( ( "Error sending command response" ) );
+        }
+    }
+}
+/*-----------------------------------------------------------*/
+
+static AzureIoTResult_t prvProcessProperties( AzureIoTHubClientPropertiesResponse_t * pxMessage,
```
hihigupt, line 400: please add some comments, Also in other places.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+        {
+            LogInfo( ( "Successfully parsed properties" ) );
+            xResult = eAzureIoTSuccess;
+        }
+    }
+
+    return xResult;
+}
+
+static void prvUpdateLocalProperties( double xNewTemperatureValue,
```
hihigupt, line 486: >t [](http://example.com/codeflow?start=3&length=1)

add 
/*---*/

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+                                      uint32_t ulPropertyVersion,
+                                      bool * xOutMaxTempChanged )
```
hihigupt, line 488: rename this pointer px<>

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+
+            break;
+
+        case eAzureIoTHubPropertiesReportedResponseMessage:
+            LogInfo( ( "Device reported property response received" ) );
+            break;
+
+        default:
+            LogError( ( "Unknown property message" ) );
+    }
```
hihigupt, line 670: this is something configAssert(false)

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+
+    LogInfo( ( "Property document payload : %.*s \r\n",
+               pxMessage->ulPayloadLength,
+               pxMessage->pvMessagePayload ) );
```
hihigupt, line 674: should this be done starting of this function?>

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Setup transport credentials.
+ */
+static uint32_t prvSetupNetworkCredentials( NetworkCredentials_t * pxNetworkCredentials )
+{
+    pxNetworkCredentials->xDisableSni = pdTRUE;
```
hihigupt, line 683: make this false

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+         * and waits for connection acknowledgment (CONNACK) packet. */
+        LogInfo( ( "Creating an MQTT connection to %s.\r\n", pucIotHubHostname ) );
+
+        xResult = AzureIoTHubClient_Connect( &xAzureIoTHubClient,
+                                             false, &xSessionPresent,
+                                             sampleazureiotCONNACK_RECV_TIMEOUT_MS );
+        configASSERT( xResult == eAzureIoTSuccess );
+
+        xResult = AzureIoTHubClient_SubscribeCloudToDeviceMessage( &xAzureIoTHubClient, prvHandleCloudMessage,
+                                                                   &xAzureIoTHubClient, sampleazureiotSUBSCRIBE_TIMEOUT );
```
hihigupt, line 803: remove this from pnp sample

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1035 @@
...
+        xResult = AzureIoTHubClient_Disconnect( &xAzureIoTHubClient );
+        configASSERT( xResult == eAzureIoTSuccess );
+
+        /* Close the network connection.  */
+        TLS_Socket_Disconnect( &xNetworkContext );
+
+        /* Wait for some time between two iterations to ensure that we do not
+         * bombard the IoT Hub. */
+        LogInfo( ( "Demo completed successfully.\r\n" ) );
+        LogInfo( ( "Short delay before starting the next iteration.... \r\n\r\n" ) );
```
hihigupt, line 860: Should we have little different flow for pnp sample... Lets disucss this

demos/CMakeLists.txt
hihigupt: I was thinking this:

we create target 
SAMPLE::AZUREIOT
SAMPLE::AZUREIOT::PNP

and in project linux/st/nxp

we do selection of target linkage using this flag. 

nit: how about naming the variable as SAMPLE_DEMO?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
hihigupt: is it possible to remove this depenency? as we have unix time function provided by port?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
hihigupt: its bit out of sync with our current formatting

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: good catch thank you

demos/CMakeLists.txt
danewalton: Ultimately I would like to just create executables for both. I don't think there's anything preventing us from creating both

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: This mimics the embedded so unless we change it there I'm going to leave as is. I think the payload is required to be JSON.

demos/CMakeLists.txt
hihigupt: That is also fine... having target for both and letting executable to add dependency.

danewalton: done

demos/projects/NXP/mimxrt1060/CMakeLists.txt
```diff
@@ -110,3 +111,24 @@ add_custom_command(TARGET ${PROJECT_NAME}
     COMMAND ${CMAKE_OBJCOPY} -Obinary $<TARGET_FILE:${PROJECT_NAME}> ${PROJECT_NAME}.bin
     COMMENT "Generate Bin file"
     VERBATIM)
+
+target_link_libraries(${PROJECT_NAME}-pnp PRIVATE
```
hihigupt, line 115: add check for this binary in ci_test.sh

https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/328fd10cd0648c25494bb35cdff1f5a3e1e03cbf/.github/scripts/ci_tests.sh#L38

for all ports

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: I don't think it's something we need to crash for.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: Yes and no. There's cases here where the hub may be sending malformed JSON in which case I don't think it's too crazy. Cases where wrong values might have been sent could maybe be handled

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: The problem is the need for all the time components. So not just `time()` but `strftime()` as well. There is a "FreeRTOS Plus Posix" library but we would then need to integrate that which is a labs project:
https://freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_POSIX/index.html

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: same as above about the labs project

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: we can but tbh I'd rather not do it now. I can make a card and get to it?
https://github.com/Azure-Samples/iot-middleware-freertos-samples/issues/52

danewalton:
As this PR stands right now, the sample passes the device certification tests:
image

demos/projects/NXP/mimxrt1060/CMakeLists.txt
```diff
@@ -1,17 +1,27 @@
...
+
+add_library(${MCUX_SDK_PROJECT_NAME})
+target_include_directories(${MCUX_SDK_PROJECT_NAME}
+  PUBLIC
+    ${BOARD_DEMO_FREERTOS_PORT_PATH}
+)
+target_link_libraries(${MCUX_SDK_PROJECT_NAME}
+  PUBLIC
+    FreeRTOS
+)
```
hihigupt, line 18: curious which part of their sdk require this depenency ?

I was under impression that our final binary requires OS depenency and middleware

demos/projects/ST/b-l475e-iot01a/CMakeLists.txt
```diff
@@ -60,3 +60,42 @@ add_custom_command(TARGET ${PROJECT_NAME}
     COMMAND ${CMAKE_OBJCOPY} -Obinary $<TARGET_FILE:${PROJECT_NAME}> ${PROJECT_NAME}.bin
     COMMENT "Generate Bin file"
     VERBATIM)
+
+# Add PnP Sample
```
hihigupt, line 64: please add this section in new samples too.

H7/l4+/windows

demos/projects/NXP/mimxrt1060/CMakeLists.txt
```diff
@@ -24,6 +34,7 @@ file(GLOB NXPCODE_SOURCES nxp_code/*.c nxp_code/lwip/*.c)
 set(PROJECT_SOURCES ${NXPCODE_SOURCES} main.c)
 
 add_executable(${PROJECT_NAME} ${PROJECT_SOURCES})
+add_executable(${PROJECT_NAME}-pnp ${PROJECT_SOURCES})
```
hihigupt, line 37: can we group them like we did for other cmakes

<simple sample>
add_executable
add_target_include
add_target_link_libraries

<pnp sample>
add_executable
add_target_include
add_target_link_libraries

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+#include "task.h"
+
+/* Azure Provisioning/IoT Hub library includes */
+#include "azure_iot_hub_client.h"
+#include "azure_iot_provisioning_client.h"
+#include "azure_iot_hub_client_properties.h"
+
+/* Azure JSON includes */
+#include "azure_iot_json_reader.h"
+#include "azure_iot_json_writer.h"
```
hihigupt, line 19: nit: sort them

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+    #error "Please define Root CA certificate of the IoT Hub(democonfigROOT_CA_PEM) in demo_config.h."
+#endif
+
+#if defined( democonfigDEVICE_SYMMETRIC_KEY ) && defined( democonfigCLIENT_CERTIFICATE_PEM )
+    #error "Please define only one auth democonfigDEVICE_SYMMETRIC_KEY or democonfigCLIENT_CERTIFICATE_PEM in demo_config.h."
+#endif
+
+#if !defined( democonfigDEVICE_SYMMETRIC_KEY ) && !defined( democonfigCLIENT_CERTIFICATE_PEM )
+    #error "Please define one auth democonfigDEVICE_SYMMETRIC_KEY or democonfigCLIENT_CERTIFICATE_PEM in demo_config.h."
+#endif
```
hihigupt, line 53: not for this PR but we should have demo_config_check.h defined and include in sample

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define sampleazureiotTRANSPORT_SEND_RECV_TIMEOUT_MS          ( 2000U )
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define sampleazureiotProvisioning_Registration_TIMEOUT_MS    ( 20U )
```
hihigupt, line 169: this is too small value ... can you updte this to atleast 3 sec... this also need to be changed in other sample too.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+    static uint32_t prvIoTHubInfoGet( NetworkCredentials_t * pXNetworkCredentials,
+                                      uint8_t ** ppucIothubHostname,
+                                      uint32_t * pulIothubHostnameLength,
+                                      uint8_t ** ppucIothubDeviceId,
+                                      uint32_t * pulIothubDeviceIdLength );
+
+#endif /* democonfigENABLE_DPS_SAMPLE */
+
+/**
+ * @brief The task used to demonstrate the MQTT API.
```
hihigupt, line 239: >e MQTT API. [](http://example.com/codeflow?start=41&length=11)

Need update

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+ *
+ * @param[in] pvParameters Parameters as passed at the time of task creation. Not
+ * used in this example.
+ */
+static void prvAzureDemoTask( void * pvParameters );
+
+/**
+ * @brief Connect to IoT Hub with reconnection retries.
```
hihigupt, line 247: >IoT Hub with reconnection retries. [](http://example.com/codeflow?start=21&length=34)

connect to endpoint with retires

same function is used for dps connect. This fix is also required in other sample

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+ * @brief Static buffer used to hold MQTT messages being sent and received.
+ */
+static uint8_t ucMQTTMessageBuffer[ democonfigNETWORK_BUFFER_SIZE ];
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Generate max min payload.
+ */
+static AzureIoTResult_t prvInvokeMaxMinCommand( AzureIoTJSONReader_t * xReader,
```
hihigupt, line 275: >xReader [](http://example.com/codeflow?start=71&length=7)

pxReader, similar comment in other places where we are using pointer to some custom types.

demos/projects/NXP/mimxrt1060/CMakeLists.txt
danewalton: https://github.com/Azure-Samples/iot-middleware-freertos-samples/runs/2933684299#step:5:3647
Here's the build failure

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+    if( xResult == eAzureIoTSuccess )
+    {
+        xResult = AzureIoTJSONReader_GetTokenString( xReader,
+                                                     ucCommandStartTimeValueBuffer,
+                                                     sizeof( ucCommandStartTimeValueBuffer ),
+                                                     &ulSinceTimeLength );
+    }
+
+    if( xResult == eAzureIoTSuccess )
+    {
```
hihigupt, line 294: how about using 

if (!success)
{
log Error fail
}
else if (!success)
{
log Error fail
}
else if ()
{
}
else
{
log Info Done processing function
}

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+               pxMessage->pvMessagePayload ) );
+
+    lCommandNameLength = strlen( sampleazureiotCOMMAND_MAX_MIN_REPORT );
+
+    if( ( lCommandNameLength == pxMessage->usCommandNameLength ) &&
+        ( strncmp( sampleazureiotCOMMAND_MAX_MIN_REPORT, pxMessage->pucCommandName, lCommandNameLength ) == 0 ) )
+    {
+        /* Is for max min report */
+        AzureIoTJSONReader_t xReader;
+        xResult = AzureIoTJSONReader_Init( &xReader, pxMessage->pvMessagePayload, pxMessage->ulPayloadLength );
```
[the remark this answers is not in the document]
hihigupt: same to all the defination used in this function

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+                }
+            }
+        }
+        else
+        {
+            LogError( ( "Error generating command payload" ) );
+
+            if( AzureIoTHubClient_SendCommandResponse( xHandle, pxMessage, 501,
+                                                       sampleazureiotCOMMAND_EMPTY_PAYLOAD,
+                                                       ulCommandPayloadLength ) != eAzureIoTSuccess )
```
hihigupt, line 395: >ulCommandPayloadLength [](http://example.com/codeflow?start=55&length=22)

ulCommandPayloadLength will have garbage value, use the strlen(sampleazureiotCOMMAND_EMPTY_PAYLOAD)/sizeof(sampleazureiotCOMMAND_EMPTY_PAYLOAD) - 1

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+                                                           ucCommandPayloadBuffer,
+                                                           ulCommandPayloadLength ) != eAzureIoTSuccess )
+                {
+                    LogError( ( "Error sending command response" ) );
+                }
+            }
+        }
+        else
+        {
+            LogError( ( "Error generating command payload" ) );
```
hihigupt, line 391: > )  [](http://example.com/codeflow?start=58&length=3)

much better if we log error code.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+                                                       ulCommandPayloadLength ) != eAzureIoTSuccess )
+            {
+                LogError( ( "Error sending command response" ) );
+            }
+        }
+    }
+    else
+    {
+        /* Not for max min report (not for this device) */
+        LogInfo( ( "Received command is not for this device" ) );
```
hihigupt, line 404: >" )  [](http://example.com/codeflow?start=59&length=4)

should we log name of command that we received.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: good catch thank you

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: did it in the other as well

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: done in both

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+    {
+        /* Reset JSON reader to the beginning */
+        xResult = AzureIoTJSONReader_Init( &xReader, pxMessage->pvMessagePayload, pxMessage->ulPayloadLength );
+        configASSERT( xResult == eAzureIoTSuccess );
+
+        while( ( xResult = AzureIoTHubClientProperties_GetNextComponentProperty( &xAzureIoTHubClient, &xReader,
+                                                                                 pxMessage->xMessageType, xPropertyType,
+                                                                                 &pucComponentName, &ulComponentNameLength ) ) == eAzureIoTSuccess )
+        {
+            xResult = AzureIoTJSONReader_TokenType( &xReader, &xTokenType );
```
hihigupt, line 451: >xResult = AzureIoTJSONReader_TokenType( &xReader, &xTokenType ); [](http://example.com/codeflow?start=12&length=64)

why do you need to know this token type, as next line irrespective of token assume token to be string token.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+                if( xTokenType != eAzureIoTJSONTokenNUMBER )
+                {
+                    LogError( ( "Invalid value for target temperature reached" ) );
+                    xResult = eAzureIoTErrorFailed;
+                    break;
+                }
+
+                /* Get desired temperature */
+                xResult = AzureIoTJSONReader_GetTokenDouble( &xReader, xOutTemperature );
+                configASSERT( xResult == eAzureIoTSuccess );
```
hihigupt, line 473: how about just using AzureIoTJSONReader_GetTokenDouble and if this fail we print the error that will give us reason why it failed.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+                xResult = AzureIoTJSONReader_NextToken( &xReader );
+                configASSERT( xResult == eAzureIoTSuccess );
+            }
+            else
+            {
+                xResult = AzureIoTJSONReader_NextToken( &xReader );
+                configASSERT( xResult == eAzureIoTSuccess );
+
+                xResult = AzureIoTJSONReader_SkipChildren( &xReader );
+                configASSERT( xResult == eAzureIoTSuccess );
```
hihigupt, line 484: we are assuming that next object is complex object.

how about using this impl:

[https://github.com/azure-rtos/netxduo/blob/ab1b3706e31f4ca9701b6c215a143386b1b655f9/addons/azure_iot/samples/sample_azure_iot_embedded_sdk_pnp.c#L328](https://github.com/azure-rtos/netxduo/blob/ab1b3706e31f4ca9701b6c215a143386b1b655f9/addons/azure_iot/samples/sample_azure_iot_embedded_sdk_pnp.c#L328)

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+    {
+        xDeviceMinimumTemperature = xDeviceCurrentTemperature;
+    }
+
+    /* Calculate the new average temperature. */
+    ulDeviceTemperatureCount++;
+    xDeviceTemperatureSummation += xDeviceCurrentTemperature;
+    xDeviceAverageTemperature = xDeviceTemperatureSummation / ulDeviceTemperatureCount;
+
+    LogDebug( ( "Client updated desired temperature variables locally." ) );
```
hihigupt, line 532: >LogDebug [](http://example.com/codeflow?start=4&length=8)

lets make it Info level logs

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: It's not so much an assumption as much as it is covering the case when it is a complex object. It's a no-op if it isn't a complex object. Same thing is done here:
https://github.com/Azure/azure-sdk-for-c/blob/75afd98a1efd41d0c1f4475f692da88547a61582/sdk/samples/iot/paho_iot_pnp_component_sample.c#L507-L509

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+
+    xResult = AzureIoTJSONWriter_AppendDouble( &xWriter, xUpdatedTemperature, sampleazureiotDOUBLE_DECIMAL_PLACE_DIGITS );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTJSONWriter_AppendEndObject( &xWriter );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    lBytesWritten = AzureIoTJSONWriter_GetBytesUsed( &xWriter );
+
+    if( lBytesWritten < 0 )
```
hihigupt, line 567: its just weird to see configAssert for all previous JSON errors, can we be consistent?

danewalton: We can. My thinking here is most of those JSON API's are relatively static. Buffer is the same size, payload is predictable, there isn't a whole lot to go wrong. For the send side it can be a little more complicated.

Do you mean for the getBytesUsed api only?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+    xResult = AzureIoTHubClientProperties_BuilderEndResponseStatus( &xAzureIoTHubClient,
+                                                                    &xWriter );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTJSONWriter_AppendEndObject( &xWriter );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    lBytesWritten = AzureIoTJSONWriter_GetBytesUsed( &xWriter );
+
+    if( lBytesWritten < 0 )
```
hihigupt, line 621: configAssert vs logging

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Property mesage callback handler
+ */
+static void prvHandleProperties( AzureIoTHubClientPropertiesResponse_t * pxMessage,
+                                 void * pvContext )
+{
+    ( void ) pvContext;
+
```
hihigupt, line 645: nit: mobe it to line 650

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+    AzureIoTResult_t xResult;
+    double xIncomingTemperature;
+    uint32_t ulVersion;
+    bool xWasMaxTemperatureChanged = false;
+
+    LogInfo( ( "Property document payload : %.*s \r\n",
```
hihigupt, line 651: >LogInfo [](http://example.com/codeflow?start=4&length=7)

this is too much noise to user, can we move it to Debug level.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+
+            if( xWasMaxTemperatureChanged )
+            {
+                prvSendNewMaxTemp( xIncomingTemperature );
+            }
+
+            break;
+
+        case eAzureIoTHubPropertiesReportedResponseMessage:
+            LogInfo( ( "Device reported property response received" ) );
```
hihigupt, line 700: >LogInfo [](http://example.com/codeflow?start=12&length=7)

Debug

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1062 @@
...
+        pxNetworkCredentials->pucPrivateKey = ( const unsigned char * ) democonfigCLIENT_PRIVATE_KEY_PEM;
+        pxNetworkCredentials->xPrivateKeySize = sizeof( democonfigCLIENT_PRIVATE_KEY_PEM );
+    #endif
+
+    return 0;
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Azure IoT demo task that gets started in the platform specific project.
```
hihigupt, line 731: add pnp scope to this doc

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
hihigupt: AzureIoTJSONWriter_GetBytesUsed

danewalton: that's fair. I'm going to take a simplifying assumption that the json builder is going to succeed and use config asserts here

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1054 @@
...
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTHubClientProperties_BuilderEndResponseStatus( &xAzureIoTHubClient,
+                                                                    &xWriter );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTJSONWriter_AppendEndObject( &xWriter );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    lBytesWritten = AzureIoTJSONWriter_GetBytesUsed( &xWriter );
```
hihigupt, line 617: is configAssert missing after this line?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
hihigupt: this change is missing

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1054 @@
...
+
+            if( xWasMaxTemperatureChanged )
+            {
+                prvSendNewMaxTemp( xIncomingTemperature );
+            }
+
+            break;
+
+        case eAzureIoTHubPropertiesWritablePropertyMessage:
+            LogInfo( ( "Device writeable property received" ) );
```
hihigupt, line 670: >LogInfo [](http://example.com/codeflow?start=12&length=7)

make this also debug

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1054 @@
...
+            }
+
+            break;
+
+        case eAzureIoTHubPropertiesReportedResponseMessage:
+            LogInfo( ( "Device reported property response received" ) );
+            break;
+
+        default:
+            LogError( ( "Unknown property message" ) );
```
hihigupt, line 695: print message type

hihigupt (approved):

demos/projects/ST/b-l475e-iot01a/CMakeLists.txt
```diff
@@ -60,3 +60,42 @@ add_custom_command(TARGET ${PROJECT_NAME}
...
+add_executable(${PROJECT_NAME}-pnp ${PROJECT_SOURCES})
+target_include_directories(${PROJECT_NAME}-pnp PUBLIC
```
hihigupt, line 66: currently in gcc_flags we set the map file to be output.map, and now we have two target generated. This will cause the later target to override previous map file.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+
+/* Device values */
+static double xDeviceCurrentTemperature = sampleazureiotDEFAULT_START_TEMP_CELSIUS;
+static double xDeviceMaximumTemperature = sampleazureiotDEFAULT_START_TEMP_CELSIUS;
+static double xDeviceMinimumTemperature = sampleazureiotDEFAULT_START_TEMP_CELSIUS;
+static double xDeviceTemperatureSummation = sampleazureiotDEFAULT_START_TEMP_CELSIUS;
+static uint32_t ulDeviceTemperatureCount = sampleazureiotDEFAULT_START_TEMP_COUNT;
+static double xDeviceAverageTemperature = sampleazureiotDEFAULT_START_TEMP_CELSIUS;
+
+/* Command Values */
```
jspaith, line 202: Nit - is it `Values` or `buffers` here?  Same feedback for the Property Values below.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+
+        xResult = prvInvokeMaxMinCommand( &xReader, &xWriter );
+
+        if( xResult == eAzureIoTSuccess )
+        {
+            ulCommandPayloadLength = AzureIoTJSONWriter_GetBytesUsed( &xWriter );
+
+            if( ulCommandPayloadLength > 0 )
+            {
+                if( AzureIoTHubClient_SendCommandResponse( pxHandle, pxMessage, 200,
```
jspaith, line 384: I recommend on `AzureIoTHubClient_SendCommandResponse` success - here and in the 501 and 404 responses below - that we have a LogInfo level that this has happened.  If a device is mysteriously not working and not able to be debugged live, logging "important" events like sample/SDK at least thinking it has sent a response can aid debug effort.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+        }
+        else
+        {
+            LogError( ( "Error generating command payload: 0x%08x", xResult ) );
+
+            if( AzureIoTHubClient_SendCommandResponse( pxHandle, pxMessage, 501,
+                                                       ( const uint8_t * ) sampleazureiotCOMMAND_EMPTY_PAYLOAD,
+                                                       strlen( sampleazureiotCOMMAND_EMPTY_PAYLOAD ) ) != eAzureIoTSuccess )
+            {
+                LogError( ( "Error sending command response" ) );
```
jspaith, line 400: On LogErrors generally, we should be capturing the return code from our function and outputting it.  We're pretty thoughtful on these errors so showing this pattern would help log based debugging.

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -512,28 +512,28 @@ static void prvAzureDemoTask( void * pvParameters )
...
+        configASSERT( xResult == eAzureIoTSuccess );
 
         #ifdef democonfigDEVICE_SYMMETRIC_KEY
             xResult = AzureIoTProvisioningClient_SetSymmetricKey( &xAzureIoTProvisioningClient,
                                                                   ( const uint8_t * ) democonfigDEVICE_SYMMETRIC_KEY,
                                                                   sizeof( democonfigDEVICE_SYMMETRIC_KEY ) - 1,
                                                                   Crypto_HMAC );
-            configASSERT( xResult == eAzureIoTProvisioningSuccess );
+            configASSERT( xResult == eAzureIoTSuccess );
         #endif // democonfigDEVICE_SYMMETRIC_KEY
```
jspaith, line 523: Nit - prexisting, but this is only use of C++ style comment in file.  We should keep completely consistent.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+    }
+    else if( ( xResult = AzureIoTJSONReader_GetTokenString( pxReader,
+                                                            ucCommandStartTimeValueBuffer,
+                                                            sizeof( ucCommandStartTimeValueBuffer ),
+                                                            &ulSinceTimeLength ) )
+             != eAzureIoTSuccess )
+    {
+        LogError( ( "Error getting token string: result 0x%08x", xResult ) );
+    }
+    else if( ( xResult = AzureIoTJSONWriter_AppendBeginObject( pxWriter ) )
```
jspaith, line 295: Nit - given we have the `/* Get the start time */` above, should we have something with like `/* Build the response buffer */` once we start building it up?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+    lCommandNameLength = strlen( sampleazureiotCOMMAND_MAX_MIN_REPORT );
+
+    if( ( lCommandNameLength == pxMessage->usCommandNameLength ) &&
+        ( strncmp( sampleazureiotCOMMAND_MAX_MIN_REPORT, ( const char * ) pxMessage->pucCommandName, lCommandNameLength ) == 0 ) )
+    {
+        /* Is for max min report */
+        xResult = AzureIoTJSONReader_Init( &xReader, pxMessage->pvMessagePayload, pxMessage->ulPayloadLength );
+        configASSERT( xResult == eAzureIoTSuccess );
+
+        xResult = AzureIoTJSONWriter_Init( &xWriter, ucCommandPayloadBuffer, sizeof( ucCommandPayloadBuffer ) );
```
jspaith, line 373: If I've never seen this pattern before - which I probably haven't if I'm using the sample - this is going to take some review for me to get my head around.

Even though not strictly necessary to someone knowing all this, I recommend a comment on this one specifically like
/* Initialize xWriter, which will be used to build the JSON response buffer, with a buffer for it to write into. */

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+            {
+                prvUpdateLocalProperties( xIncomingTemperature, ulVersion, &xWasMaxTemperatureChanged );
+                prvAckIncomingTemperature( xIncomingTemperature, ulVersion );
+            }
+            else
+            {
+                LogError( ( "There was an error processing incoming properties" ) );
+            }
+
+            if( xWasMaxTemperatureChanged )
```
jspaith, line 661: Recommend moving the `if( xWasMaxTemperatureChanged )` block up so its executed only in the `if (xResult == eAzureIoTSuccess )` block above so this will pop out as something not processed if there could have been an error.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+
+    ( void ) pvContext;
+
+    LogDebug( ( "Property document payload : %.*s \r\n",
+                pxMessage->ulPayloadLength,
+                pxMessage->pvMessagePayload ) );
+
+    switch( pxMessage->xMessageType )
+    {
+        case eAzureIoTHubPropertiesGetMessage:
```
jspaith, line 647: Again not this PR or even this repo, but I think `eAzureIoTHubPropertiesGetMessage` really means `eAzureIoTHubPropertiesGetResponseMessage`?  As currently written with just "GetMessage" it looks like server is doing a get on the device itself rather than server respending to a device initiated get.

jspaith, line 647: Are we sure this is `eAzureIoTHubClientPropertyWritable`?  This is same value we pass in block below with `eAzureIoTHubPropertiesWritablePropertyMessage` (where it's correct).  I think this is `eAzureIoTHubPropertiesGetMessage` ??

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+
+            if( xWasMaxTemperatureChanged )
+            {
+                prvSendNewMaxTemp( xIncomingTemperature );
+            }
+
+            break;
+
+        case eAzureIoTHubPropertiesWritablePropertyMessage:
+            LogDebug( ( "Device writeable property received" ) );
```
jspaith, line 669: Feels like a fair amount of dup here and in the `eAzureIoTHubPropertiesGetMessage` above, could we move out to a helper function that both of these call?

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+    *pxOutTemperature = 0.0;
+
+    xResult = AzureIoTJSONReader_Init( &xReader, pxMessage->pvMessagePayload, pxMessage->ulPayloadLength );
+    configASSERT( xResult == eAzureIoTSuccess );
+
+    xResult = AzureIoTHubClientProperties_GetPropertiesVersion( &xAzureIoTHubClient, &xReader, pxMessage->xMessageType, ulOutVersion );
+
+    if( xResult != eAzureIoTSuccess )
+    {
+        LogError( ( "Error getting the property version" ) );
```
jspaith, line 443: I won't put everywhere as that'd be tedious for all of us, but this is another place that we should be outputting the error code of result in log.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+    }
+    else
+    {
+        /* Reset JSON reader to the beginning */
+        xResult = AzureIoTJSONReader_Init( &xReader, pxMessage->pvMessagePayload, pxMessage->ulPayloadLength );
+        configASSERT( xResult == eAzureIoTSuccess );
+
+        while( ( xResult = AzureIoTHubClientProperties_GetNextComponentProperty( &xAzureIoTHubClient, &xReader,
```
jspaith, line 451: What happens if a non-NULL componentName comes back from this?  We should log a message and continue processing otherwise.  Is is, if there was a `sampleazureiotPROPERTY_TARGET_TEMPERATURE_TEXT` but on a component we'd process which we don't want.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+                    LogError( ( "Error getting the property version" ) );
+                    break;
+                }
+
+                xResult = AzureIoTJSONReader_NextToken( &xReader );
+                configASSERT( xResult == eAzureIoTSuccess );
+            }
+            else
+            {
+                xResult = AzureIoTJSONReader_NextToken( &xReader );
```
jspaith, line 476: We should (a) log a message indicating an unsupported property arrived.  And (b) put in comments that we have to skip past that property's JSON in the calls below so we're positioned for next read through the loop.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+/**
+ * @brief Send updated maximum temperature value to IoT Hub.
+ */
+static void prvSendNewMaxTemp( double xUpdatedTemperature )
+{
+    AzureIoTResult_t xResult;
+    AzureIoTJSONWriter_t xWriter;
+    int32_t lBytesWritten;
+
+    xResult = AzureIoTJSONWriter_Init( &xWriter, ucPropertyPayloadBuffer, sizeof( ucPropertyPayloadBuffer ) );
```
jspaith, line 545: Again per my comments above on other JSONWriter_Init, recomend some commentary around this.

jspaith, line 545: Nit - This is also one where a comment like /* Building the new temperature buffer */ here and then later on, right before the `AzureIoTHubClient_SendPropertiesReported` call, something like /* Sending the temperature to endpoint */ might make this a bit more clear.

jspaith: Similar feedback to consider with `prvAckIncomingTemperature` and maybe other places?

jspaith (changes requested):
Marking request mostly because I'm not I think we need to look again at the "Are we sure this is `eAzureIoTHubClientPropertyWritable` comment down below since that's impact functionality potentially.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: added here and elsewhere in the file

demos/sample_azure_iot/sample_azure_iot.c
danewalton: yup ran the uncrustify on it and it picked it up

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+    AzureIoTJSONWriter_t xWriter;
+    int32_t lCommandNameLength;
+    int32_t ulCommandPayloadLength;
+    AzureIoTHubClient_t * pxHandle = ( AzureIoTHubClient_t * ) pvContext;
+
+    LogInfo( ( "Command payload : %.*s \r\n",
+               pxMessage->ulPayloadLength,
+               pxMessage->pvMessagePayload ) );
+
+    lCommandNameLength = strlen( sampleazureiotCOMMAND_MAX_MIN_REPORT );
```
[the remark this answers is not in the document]
danewalton: Himanshu had the some comment. I can change but I remember reading somewhere that compilers will convert strlen calls on string literals at compile time to integers. We confirmed with gcc, clang, and arm-gcc that calls to `strlen` with string literals is automatically converted to the number at compile time.

MicrosoftTeams-image (5)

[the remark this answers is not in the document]
hihigupt: I agree .. but this is confusing and very much depend upon smartness of compiler.

Confusion mostly coming from by the definition of the operation:

sizeof() gives you the size of the data type, not the size of a particular instance of that type in memory.
strlen() gives you size of memory in runtime.

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -0,0 +1,1051 @@
...
+    {
+        LogError( ( "Error getting token string: result 0x%08x", xResult ) );
+    }
+    else if( ( xResult = AzureIoTJSONWriter_AppendBeginObject( pxWriter ) )
+             != eAzureIoTSuccess )
+    {
+        LogError( ( "Error appending begin object: result 0x%08x", xResult ) );
+    }
+    else if( ( xResult = AzureIoTJSONWriter_AppendPropertyWithDoubleValue( pxWriter, ( const uint8_t * ) sampleazureiotCOMMAND_MAX_TEMP,
+                                                                           strlen( sampleazureiotCOMMAND_MAX_TEMP ),
```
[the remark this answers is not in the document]
danewalton: commented above

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: good call I'll look for lacking places

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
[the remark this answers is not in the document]
danewalton: that's fine I can update

demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
danewalton: I think this is right. This mirrors this section of the embedded:
https://github.com/Azure/azure-sdk-for-c/blob/97cde2c6f5d4fc0b411c8ea9b61192f6f9424a0d/sdk/samples/iot/paho_iot_pnp_component_sample.c#L529-L554

There's a chance we need some naming massaging here because these are the two current types we might be confusing here:

`AzureIoTHubMessageType_t`
https://github.com/Azure/azure-iot-middleware-freertos/blob/99443168dc24d431edc44cd253635996d429e433/source/include/azure_iot_hub_client.h#L61-L68

 `AzureIoTHubClientPropertyType_t`
https://github.com/Azure/azure-iot-middleware-freertos/blob/99443168dc24d431edc44cd253635996d429e433/source/include/azure_iot_hub_client_properties.h#L221-L227

In the embedded we also use the "writeable" property type as well.

danewalton: We can discuss whether there's some name improvements regardless?

jspaith: OK I think I see - the `pxMessage->xMessageType` is indeed making it down to lower layer when `prvProcessProperties` passes in `pxMessage->xMessageType` to `AzureIoTHubClientProperties_GetNextComponentProperty`.

Naming is ongoing discussion I'm all for having but not right before pub preview.

jspaith (approved):

danewalton:
image

Just to dot I's and cross T's. As the code is currently in this PR, it passed the PnP certification tests

PR #63, opened 2021-07-13
add remaining boards to gitignore

danewalton:


.gitignore
```diff
@@ -353,4 +353,7 @@ MigrationBackup/
 /build
 /mimxrt1060
 /build_linux
+/build_windows
 /b-l475e-iot01a
```
hihigupt, line 357: how about adding these folders


https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/d0e25bbeba094b315a0b808fa825abbabcfc59f8/.github/scripts/ci_tests.sh#L41

build_pc_linux
build_st_b-l475e-iot01a
build_st_stm32h745i-disco
build_st_b-l4s5i-iot01a
build_nxp_mimxrt1060

and making out .vscode also to generate same folder as checkin.. or other way around ... Just want to make sure we are use same folders in .vscode and checkin

ewertons (approved):

.gitignore
```diff
@@ -353,4 +353,14 @@ MigrationBackup/
...
+/b-l4s5i-iot01a
+/stm32h745i-disco
+
+# Gate build directories
+build_pc_linux
```
hihigupt, line 362: is there reason why we have absolute path for above and relative for these?

danewalton: ah nope bad copy paste

hihigupt (approved):

PR #71, opened 2021-07-19
Fix map file override

hihigupt:
https://github.com/Azure-Samples/iot-middleware-freertos-samples/issues/61

### DIFF
--- a/demos/projects/PC/windows/msvc_flags.cmake
+++ b/demos/projects/PC/windows/msvc_flags.cmake
@@ -10,3 +10,7 @@ add_compile_options(
       $<$<CONFIG:Release>:/MT> #--|
       ${MCU_C_FLAGS})
 add_link_options(/NODEFAULTLIB:libcmtd.lib)
+
+function(add_map_file TARGET_NAME MAP_FILE_NAME)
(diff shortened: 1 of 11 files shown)


demos/projects/PC/windows/msvc_flags.cmake
```diff
@@ -10,3 +10,7 @@ add_compile_options(
       $<$<CONFIG:Release>:/MT> #--|
       ${MCU_C_FLAGS})
 add_link_options(/NODEFAULTLIB:libcmtd.lib)
+
+function(add_map_file TARGET_NAME MAP_FILE_NAME)
+    target_link_options(${TARGET_NAME} PRIVATE /MAP)
+endfunction()
```
danewalton, line 16: nit: can we put this in a top level `.cmake` file that we include once and only have one instance of?

danewalton (approved):

demos/projects/PC/windows/msvc_flags.cmake
hihigupt: this function needs to be aware of compiler .... right now compiler aware file is inside each project as compiler_flags.cmake (e.g. gcc_flags.cmake). To keep it inline with the existing structure I added to compiler aware specfic cmake.

PR #72, opened 2021-07-19
Fix project readme

hihigupt:


demos/projects/PC/linux/README.md
```diff
@@ -84,7 +84,7 @@ Execute the command below to find which index you got for the ``rtosveth1`` (ind
 ```
 Look for line #138 in `FreeRTOSConfig.h` and update `configNETWORK_INTERFACE_TO_USE` with the number you got in the previous step.
 
-**Example**: if you got ``4.rtosveth1 [Up, Running]`` in the previous step, you'll update line #138 to look like this ``#define configNETWORK_INTERFACE_TO_USE ( 4L )``
+**Example**: if you got ``4.rtosveth1 [Up, Running]`` in the previous step, you'll  update macro ``configNETWORK_INTERFACE_TO_USE`` to look like this ``#define configNETWORK_INTERFACE_TO_USE ( 4L )``
```
jspaith, line 87: Nit - diff looks like extra white space for `you'll  update`, just one space needed.

jspaith: Another nit, maybe dumb MD comment, but why are there two tick marks in front of configNETWORK_INTERFACE_TO_USE.  This was pre-existing but I thought there should just be one??

jspaith (approved):

demos/projects/PC/linux/README.md
hihigupt: extra careful use of back ticks, but not consistent, so will move them back to single tick.

"If the word or phrase you want to denote as code includes one or more backticks, you can escape it by enclosing the word or phrase in double backticks"

danewalton (approved):

PR #74, opened 2021-07-21
Fix NXP sample cmake

hihigupt:
When using NXP SDK as static library, sample binaries failed to
flash to board. This was happening because boot headers were missing
from final sample binary. Drop to these boot headers is done by linker when
boot headers are part of static library. Following stack overflow
describes the problem:
https://stackoverflow.com/questions/44675582/gcc-how-to-tell-linker-not-to-skip-unused-sections

As part of this fix adding boot header object files to linking step.


wduraes (approved):
both PnP and non-PnP samples worked fine on my NXP device.

ewertons (approved):

PR #128, opened 2021-11-30
Add information about known issue workaround

wduraes:
Adding an note to point users to the ESP-IDF CCache issue workaround


ewertons (approved):
Ideally it should be a table so we can add more items, but for this first version it should be more than enough.

PR #135, opened 2022-01-18
fix sample

felixkroemer:
## Purpose

The span for the properties is initialized with the size of the property bag instead of the size of the buffer.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


danewalton:
Good catch @felixkroemer 
Thanks!

danewalton (approved):

PR #139, opened 2022-01-31
Update root certs and mbedTLS api to add root certs

danewalton:
The previous PEM method of adding certs wasn't working for all scenarios. This adds them in byte form.

Also changes the mbedtls api to use `mbedtls_x509_crt_parse_der_nocopy()` so that it doesn't have to create a copy of the previous string (uses the cert array as a static array).


momuno (approved):
assuming it works, looks good! :D

CIPop:
A few comments:
1. Please add instructions on what these arrays are and how to generate them (i.e. xxd or other tools).
2. Are they base64 encoded? If yes, then can they be DER binary instead to save space?
3. Out of scope likely but we need to make sure customers do not copy-paste this implementation. We should have proper documentation like in C-SDK to keep CA certificates in a proper CA store. This is different on different TLS stacks and devices. (E.g. Espressif's x509 certificate bundle: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_crt_bundle.html).

danewalton:
> A few comments:
> 
> 1. Please add instructions on what these arrays are and how to generate them (i.e. xxd or other tools).
> 2. Are they base64 encoded? If yes, then can they be DER binary instead to save space?
> 3. Out of scope likely but we need to make sure customers do not copy-paste this implementation. We should have proper documentation like in C-SDK to keep CA certificates in a proper CA store. This is different on different TLS stacks and devices. (E.g. Espressif's x509 certificate bundle: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_crt_bundle.html).

https://github.com/Azure-Samples/iot-middleware-freertos-samples/pull/140
Added for additional notice about certificates.

PR #140, opened 2022-02-03
Add certificate notice to samples

danewalton:


demos/projects/ESPRESSIF/aziotkit/config/demo_config.h
```diff
@@ -127,6 +127,12 @@
  *  - DigiCert Global Root G2 
  *  - Microsoft RSA Root Certificate Authority 2017
  *
+ * @warning Hard coding certificates is not recommended by Microsoft as a best
+ * practice for production scenarios. Please see our document here for notes on best practices.
+ * https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/main/docs/certificate-notice.md
+ *
+ * These certs are DER formats of the above listed certificates.
```
ewertons, line 134: Not sure if we should include an explicit call to an openssl function here. What about just describing the need for DER and documenting how to get it in the certificate-notice.md ? Or somewhere else other than the source code/headers.

Same question for other places below.

danewalton: That mainly came from Chris's comment here:
https://github.com/Azure-Samples/iot-middleware-freertos-samples/pull/139#issuecomment-1026381756

I don't mind if they are or aren't here.

ewertons (approved):

demos/projects/ESPRESSIF/aziotkit/config/demo_config.h
CIPop: Maybe we can add all certificate documentation in the MD file simplifying the comment to the warning only?

docs/certificate-notice.md
```diff
@@ -0,0 +1,12 @@
+# Certificate Notice
+
+__IMPORTANT:__
+
+- The content of this repository is not guaranteed to be up to date. It is also not guaranteed to be a complete list of CA certificates used across Azure Clouds or when using Azure Stack, Azure IoT Edge, or Azure Protocol Gateway.
```
CIPop, line 5: I feel that the context hasn't been set for the first sentence. Maybe leaving just the second part is sufficient?

docs/certificate-notice.md
```diff
@@ -0,0 +1,12 @@
...
+- Always prefer using the local system's Trusted Root Certificate Authority store instead of hardcoded certificates (e.g. the hardcoded certificates used in our `demo_config.h` files).
+- Azure Root certificates may change with or without prior notice (e.g., if they expire or are revoked). It is important that devices are able to add or remove trusted root certificates.
+- Support for at least two certificates is required to maintain device connectivity during CA certificate changes.
+
+## Additional Information
```
CIPop, line 10: We could add the DER information from the comments here together with an example on how a Base64 encoded certificate can be transformed into a DER here.

CIPop (approved):

docs/certificate-notice.md
danewalton: I think that gets in the weeds a little too much. I'll probably leave it as is to not have this readme be an openssl and certificates lesson.

PR #156, opened 2022-03-30
remove non-relevant compilations for OTA

danewalton:
Removes compilations for boards which aren't used for OTA right now.


danewalton:
This does bring up an interesting compilation hiccup. If the middleware is compiled as one static library, ADU takes a dependency on an http port header. This would require someone to provide one even if they don't use ADU. We could either provide a dummy header to satisfy the compilation and break at runtime, or attempt to separate out ADU as a separate static lib with a different name. The latter seems like it might be the better option.

PR #167, opened 2022-04-22
Add x509 Cert capability to esp32

danewalton:


demos/projects/ESPRESSIF/aziotkit/README.md
```diff
@@ -77,9 +77,36 @@ Parameter | Value
...
+```txt
+# PEM Formatted (WRONG)
+-----BEGIN CERTIFICATE-----
+MIIBJDCBywIUfeHrebBVa2eZAbouBgACp9R3BncwCgYIKoZIzj0EAwIwETEPMA0G
+...
+vTfQahwsxN3xink9z1gtirrjQlqDAiEAyU+6TUJcG6d9JF+uJqsLFpsbbF3IzGAw
+yC+koNRC0MU=
+-----END CERTIFICATE-----
+
+# Single Line (CORRECT)
```
ewertons, line 106: Suggestion: show the correct way of doing it right at the top first, and then the comment "Note that the certi..." and finally the wrong example. Seems more intuitive like that.

demos/projects/ESPRESSIF/esp32/components/sample-azure-iot/CMakeLists.txt
```diff
@@ -15,6 +15,14 @@ else()
     )
 endif()
 
+# Correctly format the cert and key if relevant
```
ewertons, line 18:
```suggestion
# kconfig does not support multiline strings.
# For certificates, we use as workaround escaping the newlines
# in certificate and key so they can be entered as a single
# string in kconfig.
# The routine below unescapes the newlines so the values
#  can be correctly interpreted by the code.
```

demos/projects/ESPRESSIF/aziotkit/components/sample-azure-iot/CMakeLists.txt
```diff
@@ -5,6 +5,14 @@ set(ROOT_PATH
     ${CMAKE_CURRENT_LIST_DIR}/../../../../../..
 )
 
+# Correctly format the cert and key if relevant
```
ewertons, line 8:
```suggestion
# kconfig does not support multiline strings.
# For certificates, we use as workaround escaping the newlines
# in certificate and key so they can be entered as a single
# string in kconfig.
# The routine below unescapes the newlines so the values
#  can be correctly interpreted by the code.
```

ewertons (approved):
Approved with suggestions.

PR #171, opened 2022-05-03
Add debug and release build versions

### LINKED ISSUE #141: Show memory usage with each build
Proposed enhancement: for all samples or at least those using ARM GCC we could show memory usage with each build using `-print-memory-usage`.
Example diff for `gcc_flags.cmake` for b-l475e-iot01a:

```
set(CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS} "-Wl,--gc-sections,-print-memory-usage \
    -static -z muldefs -mthumb -Wl,--start-group -lc -lm -Wl,--end-group")
```

danewalton:
- Adds debug and release builds.
- Adds memory usage output for boards.

closes #141

### DIFF
--- a/.github/scripts/ci_tests.sh
+++ b/.github/scripts/ci_tests.sh
@@ -39,7 +39,7 @@ function sample_build() {
     then
       idf.py build -DCMAKE_BUILD_TYPE=$buildver -C ./demos/projects/ESPRESSIF/$board
       echo -e "::group::Print Size for $board $buildver"
-      ninja -C ./demos/projects/ESPRESSIF/$board size-components
+      ninja -C ./demos/projects/ESPRESSIF/$board/build size-components
     else
       cmake -G Ninja -DBOARD=$board -DVENDOR=$vendor -B$outdir -DFREERTOS_PATH=$TEST_FREERTOS_SRC -DCMAKE_BUILD_TYPE=$buildver .
       cmake --build $outdir > build.txt
--- a/.github/scripts/ci_tests.sh
+++ b/.github/scripts/ci_tests.sh
@@ -35,18 +35,11 @@ function sample_build() {
     local outdir=$3
     local buildver=$4
 
-    if [ -z "$buildver" ]
-    then
-      buildver="Debug"
-    fi
-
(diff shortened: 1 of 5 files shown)


.github/scripts/ci_tests.sh
```diff
@@ -56,44 +66,119 @@ for arg in $RUN_BOARDS_BUILD
...
 
-            echo -e "::group::Building sample for NXP mimxrt1060 port"
-            sample_build "NXP" "mimxrt1060" "build_nxp_mimxrt1060"
+            echo -e "::group::Building sample for NXP mimxrt1060 port - Debug"
+            sample_build "NXP" "mimxrt1060" "build_nxp_mimxrt1060" "Debug"
             exit_if_binary_does_not_exist "build_nxp_mimxrt1060" "iot-middleware-sample.elf"
             exit_if_binary_does_not_exist "build_nxp_mimxrt1060" "iot-middleware-sample-pnp.elf"
+
+            echo -e "::group::Print Size for Debug"
+            grep -A9 "Linking C executable demos/projects/NXP/mimxrt1060/iot-middleware-sample.elf" build.txt
```
jspaith, line 92: Nit - since you're intending this as a build output helper only the grep can work, though consider greping on "memory region" which seems less likely to change than the "linking..." always being before size in future linker output changes.

jspaith (approved):

.github/scripts/ci_tests.sh
danewalton: Yea I hear that. If the linker output changes, this grep should then fail and the build will fail so we can likely resolve it then.

.github/scripts/ci_tests.sh
```diff
@@ -33,19 +33,29 @@ function sample_build() {
     local vendor=$1
     local board=$2
     local outdir=$3
+    local buildver=$4
+
+    if [ -z "$buildver" ]
+    then
+      buildver="Debug"
```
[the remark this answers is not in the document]
danewalton: Changed to require it to be specified.

PR #179, opened 2022-06-03
Add JWS Validation

danewalton:
Adds JWS validation for the manifest using mbedTLS.

- Note that many logs in the source code are commented out for now. They are useful if we need to debug and I added a TODO to remove them in the future.
- Below is a screenshot of the logs as it is verified:

image


demos/common/utilities/jws.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ */
+
+#ifndef JWS_H
+#define JWS_H
+
+#include <stdint.h>
+
+/**
+ * @brief Verify the manifest from ADU.
```
ericwol-msft, line 15:
```suggestion
 * @brief Authenticate the manifest from ADU.
```

demos/common/utilities/jws.h
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @param pucManifest The escaped manifest from the ADU twin property.
+ * @param ulManifestLength The length of \p pucManifest.
+ * @param pucJWS The JWS used to verify \p pucManifest.
+ * @param ulJWSLength The length of \p pucJWS.
+ * @return uint32_t
```
ericwol-msft, line 21: what is this value?

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+
+    AzureIoTResult_t xResult = prvSplitJWS( ( unsigned char * ) pucJWS, ulJWSLength,
+                                            &pucHeader, &ulHeaderLength,
+                                            &pucPayload, &ulPayloadLength,
+                                            &pucSignature, &ulSignatureLength );
+    prvSwapToUrlEncodingChars( pucSignature, ulSignatureLength );
+
+    LogInfo( ( "---JWS Base64 Decode Header---\n" ) );
+    int32_t outJWSDecodedHeaderLength;
+    lMbedResult = mbedtls_base64_decode( ( unsigned char * ) ucBase64DecodedHeader, sizeof( ucBase64DecodedHeader ), ( size_t * ) &outJWSDecodedHeaderLength, ( const unsigned char * ) pucHeader, ulHeaderLength );
```
ericwol-msft, line 303: ugh. after decoding ucBase64DecodedHeader is nothing special, just a buffer, while is actually encoded. I would swap the tag around as in:

pucHeader -> pucBase64EncodedHeader
ucBase64DecodedHeader ->ucHeader

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+    /* TODO: here and elsewhere, remove verbose logs */
+    /* LogInfo(( "\tmbedTLS Return: 0x%x\n", lMbedResult )); */
+    /* LogInfo(( "\tOut Decoded Size: %i\n", outJWSDecodedHeaderLength )); */
+    /* LogInfo(( "\t%.*s\n\n", ( int ) outJWSDecodedHeaderLength, ( char * ) ucBase64DecodedHeader )); */
+
+    LogInfo( ( "---JWS Base64 Decode Payload---\n" ) );
+    int32_t outJWSDecodedLength;
+    lMbedResult = mbedtls_base64_decode( ( unsigned char * ) ucBase64DecodedPayload, sizeof( ucBase64DecodedPayload ), ( size_t * ) &outJWSDecodedLength, ( const unsigned char * ) pucPayload, ulPayloadLength );
```
ericwol-msft, line 311: ucBase64DecodedPayload -> ucPayload
pucPayload -> pucBase64EncodedPayload

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+    /* LogInfo(( "\tmbedTLS Return: 0x%x\n", lMbedResult )); */
+    /* LogInfo(( "\tOut Decoded Size: %i\n", outJWSDecodedLength )); */
+    /* LogInfo(( "\t%.*s\n\n", ( int ) outJWSDecodedLength, ( char * ) ucBase64DecodedPayload )); */
+
+    LogInfo( ( "---JWS Base64 Decode Signature---\n" ) );
+    int32_t outJWSDecodedSignatureLength;
+    lMbedResult = mbedtls_base64_decode( ( unsigned char * ) ucBase64DecodedSignature, sizeof( ucBase64DecodedSignature ), ( size_t * ) &outJWSDecodedSignatureLength, ( const unsigned char * ) pucSignature, ulSignatureLength );
```
ericwol-msft, line 318: ucBase64DecodedSignature -> ucSignature
pucSignature -> pucBase64EncodedSignature

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+
+    /*------------------- Parse JSK JSON Payload ------------------------*/
+
+    /* The "sjwk" is the signed signing public key */
+    LogInfo( ( "---Parsing JWS JSON Payload---\n" ) );
+    unsigned char * pucJWKManifest;
+    az_span xJWKManifestSpan;
+
+    /*TODO: REMOVE THIS HACK */
+    ucBase64DecodedHeader[ outJWSDecodedHeaderLength ] = '"';
```
ericwol-msft, line 332: assuming outJWSDecodedHeaderLength is < sizeof(ucBase64DecodedHeader) - 2

ericwol-msft: for people playing along, why u doing this?

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+            return _xAzResult;                       \
+        }                                            \
+    } while( 0 )
+
+#define jwsRSA3072_SIZE            384
+#define jwsSHA256_SIZE             32
+#define jwsPKCS7_PAYLOAD_OFFSET    19
+
+#define jwsSHA256_JSON_VALUE       "sha256"
+#define jwsSJWK_JSON_VALUE         "sjwk"
```
ericwol-msft, line 41: can these be const uint8_t* jwsSJWK_JSON_VALUE = "sjwk";

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+        if( AzureIoTJSONReader_TokenIsTextEqual( &xJSONReader, ( const uint8_t * ) jwsSJWK_JSON_VALUE, sizeof( jwsSJWK_JSON_VALUE ) - 1 ) )
+        {
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( &xJSONReader ) );
+            break;
+        }
+        else
+        {
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( &xJSONReader ) );
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_SkipChildren( &xJSONReader ) );
+            xResult = AzureIoTJSONReader_NextToken( &xJSONReader );
```
ericwol-msft, line 349: what if this fails

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+
+    if( az_span_is_content_equal( rootKeyIDSpan, kidSpan ) )
+    {
+        LogInfo( ( "Using the correct root key\n" ) );
+    }
+    else
+    {
+        LogInfo( ( "Using the wrong root key\n" ) );
+
+        while( 1 )
```
ericwol-msft, line 437: ????

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+
+    az_span rootKeyIDSpan = az_span_create( ( uint8_t * ) AzureIoTADURootKeyId, sizeof( AzureIoTADURootKeyId ) - 1 );
+
+    if( az_span_is_content_equal( rootKeyIDSpan, kidSpan ) )
+    {
+        LogInfo( ( "Using the correct root key\n" ) );
+    }
+    else
+    {
+        LogInfo( ( "Using the wrong root key\n" ) );
```
ericwol-msft, line 435:
```suggestion
        LogError( ( "Using the wrong root key\n" ) );
```

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+    az_span nSpan = AZ_SPAN_EMPTY;
+    az_span eSpan = AZ_SPAN_EMPTY;
+    az_span algSpan = AZ_SPAN_EMPTY;
+    LogInfo( ( "--- Parse Signing Key Payload ---\n" ) );
+
+    AzureIoTJSONReader_Init( &xJSONReader, ( const uint8_t * ) ucBase64DecodedJWKPayload, outDecodedJWKSizeTwo );
+    /*Begin object */
+    azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( &xJSONReader ) );
+    /*Property Name */
+    azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( &xJSONReader ) );
```
ericwol-msft, line 453: should we not validate that prop name is "jwk"?

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+
+    while( xResult == eAzureIoTSuccess && ( az_span_size( nSpan ) == 0 || az_span_size( eSpan ) == 0 || az_span_size( algSpan ) == 0 ) )
```
ericwol-msft, line 455: so what happens if xResult is in a failed state when exiting the loop?

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+
+    /*------------------- Verify that the SHAs match ------------------------*/
+    ulVerificationResult = prvJWS_SHA256Calculate( ( const unsigned char * ) pucManifest,
+                                                   ulManifestLength,
+                                                   ucEscapedManifestSHACalculation );
+
+    if( ulVerificationResult != 0 )
+    {
+        LogError( ( "SHA256 Calculation failed" ) );
+    }
```
ericwol-msft, line 541: return ulVerificationResult; ????

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,593 @@
...
+ * @param pucE The exponent used for the key.
+ * @param ulELength The length of \p pucE.
+ * @param pucBuffer The buffer used as scratch space to make the calculations. It should be at least
+ * `jwsRSA3072_SIZE` + `jwsSHA256_SIZE` in size.
+ * @param ulBufferLength The length of \p pucBuffer.
+ * @return uint32_t The result of the operation.
+ * @retval 0 if successful.
+ * @retval Non-0 if not successful.
+ */
+static uint32_t prvJWS_RS256Verify( unsigned char * pucInput,
```
ericwol-msft, line 183: wounder if this can be broken down at each section instead of one big function.

demos/projects/ESPRESSIF/aduotaupdate/main/azure_iot_freertos_esp32_pnp_model.c
```diff
@@ -255,6 +257,17 @@ void vHandleWritableProperties( AzureIoTHubClientPropertiesResponse_t * pxMessag
...
             }
 
+            LogInfo( ( "Verifying JWS Manifest" ) );
+            uint32_t ulJWSVerify = JWS_ManifestVerify( (char*)xAzureIoTAduOtaUpdateRequest.pucUpdateManifest,
+                                                       xAzureIoTAduOtaUpdateRequest.ulUpdateManifestLength,
+                                                       (char*)xAzureIoTAduOtaUpdateRequest.pucUpdateManifestSignature,
+                                                       xAzureIoTAduOtaUpdateRequest.ulUpdateManifestSignatureLength);
+            if (ulJWSVerify != 0)
+            {
+              LogError( ( "JWS_ManifestVerify failed: was not validated successfully" ) );
```
ericwol-msft, line 267:
```suggestion
              LogError( ( "JWS_ManifestVerify failed: JWS was not validated successfully" ) );
```

demos/common/utilities/jws_using_mbedtls.c
ericwol-msft: might make unit tests easier

demos/common/utilities/jws_using_mbedtls.c
danewalton: ahhh thanks

demos/common/utilities/jws_using_mbedtls.c
danewalton: they can but having it this way allows for a `sizeof()` call

demos/projects/ESPRESSIF/adu/main/azure_iot_freertos_esp32_pnp_model.c
```diff
@@ -254,6 +256,17 @@ void vHandleWritableProperties( AzureIoTHubClientPropertiesResponse_t * pxMessag
                 return;
             }
 
+            LogInfo( ( "Verifying JWS Manifest" ) );
+            uint32_t ulJWSVerify = JWS_ManifestAuthenticate( (char*)xAzureIoTAduUpdateRequest.pucUpdateManifest,
```
jspaith, line 260: Since the parameter to function is const, shouldn't the typecast be (const char*).  Though please check out my feedback about this in `JWS_ManifestAuthenticate` header before changing as I'm not sure char* is right.

demos/common/utilities/jws.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @param pucManifest The escaped manifest from the ADU twin property.
+ * @param ulManifestLength The length of \p pucManifest.
+ * @param pucJWS The JWS used to authenticate \p pucManifest.
+ * @param ulJWSLength The length of \p pucJWS.
+ * @return uint32_t The return value of this function.
+ * @retval 0 if successful.
+ * @retval Otherwise if failed.
+ */
+uint32_t JWS_ManifestAuthenticate( const char * pucManifest,
```
jspaith, line 27: `const char*` implies a null termination char.  Since you're passing in length anyway shouldn't this be like uint8_t (or whatever it is?).

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+#define jwsPKCS7_PAYLOAD_OFFSET    19
+
+#define jwsSHA256_JSON_VALUE       "sha256"
+#define jwsSJWK_JSON_VALUE         "sjwk"
+#define jwsKID_JSON_VALUE          "kid"
+#define jwsN_JSON_VALUE            "n"
+#define jwsE_JSON_VALUE            "e"
+#define jwsALG_JSON_VALUE          "alg"
+
+static unsigned char ucJWSHeader[ 1400 ];
```
jspaith, line 47: Where does the number 1400 come from?  And these #'s generally?  Could we link to a spec someplace with like limits or JWS spec or ...?

jspaith, line 47: I'm worried that this data is always going to be on the .bss, even when the vast majority of the time it's not actually going to be used.  And it's about 2KB tax here.

I get this complicates the API, but would we be better off having the application to pass in a scratch buffer (and requiring it be some size, also passed in, we could have a "define REQUIRED_SIZE <...>" that they could just use to know right thing ).  

In this case they could decide whether they'd want to allocate all this off the stack (so it'd be temporary) or heap or wherever made most sense for them. 

Vaguely I recall Himanshu did something similar to avoid a malloc() in product code.

jspaith: Another advantage of passing in is that this code as-is isn't thread safe.  I get it's unlikely multiple ADU's would happen at same time, unless... there was an ADU update for a device identity and module going simultaneously.  (Maybe the module represents some peripheral or something that's tied into ADU).  If someone had a scheme like that, it's not inconceivable that on a reconnect then the app would fire ADU off for both module+device identities right away and you could get into this state.

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+                                            &pucBase64EncodedSignature, &ulBase64SignatureLength );
+    prvSwapToUrlEncodingChars( pucBase64EncodedSignature, ulBase64SignatureLength );
+
+    /* Note that we do not use mbedTLS to base64 decode values since we need the ability to assume padding characters. */
+    /* mbedTLS will stop the decoding short and we would then need to hack in the remaining characters. */
+    LogInfo( ( "---JWS Base64 Decode Header---\n" ) );
+    int32_t outJWSHeaderLength;
+    az_span xJWSBase64EncodedHeaderSpan = az_span_create( pucBase64EncodedHeader, ulBase64EncodedHeaderLength );
+    az_span xJWSHeaderSpan = az_span_create( ucJWSHeader, sizeof( ucJWSHeader ) );
+    az_result xCoreResult = az_base64_decode( xJWSHeaderSpan, xJWSBase64EncodedHeaderSpan, &outJWSHeaderLength );
```
jspaith, line 317: Need to check error code on this (and other places).

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+    /* TODO: here and elsewhere, remove verbose logs */
+    /* LogInfo(( "az_base64_decode return: 0x%x\n", xCoreResult)); */
+    /* LogInfo(( "\tOut Decoded Size: %i\n", outJWSHeaderLength )); */
+    /* LogInfo(( "\t%.*s\n\n", ( int ) outJWSHeaderLength, ( char * ) ucJWSHeader )); */
+
+    LogInfo( ( "---JWS Base64 Decode Payload---\n" ) );
+    int32_t outJWSPayloadLength;
+    az_span xJWSBase64EncodedPayloadSpan = az_span_create( pucBase64EncodedPayload, ulBase64EncodedPayloadLength );
+    az_span xJWSPayloadSpan = az_span_create( ucJWSPayload, sizeof( ucJWSPayload ) );
```
jspaith, line 326: First you should checkout comments I have about whether to use statics for stuf8f like `ucJWSPayload` at top of this file.

That said, we do this parse quite early in the process and only start using the buffer like 200 lines later, like line 578.  Could we potentially only actually do unencode when we need to?  The advantage of this is that we could have a single scratch buffer, so we only do the base64_decode and then immediately do JSON parsing on it.  And at other places we need the scratch, we do the calculation in that and then we're done.  So we end up needing less memory than having each buffer allocated on its own separately.

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+
+    AzureIoTResult_t xResult = prvSplitJWS( ( unsigned char * ) pucJWS, ulJWSLength,
+                                            &pucBase64EncodedHeader, &ulBase64EncodedHeaderLength,
+                                            &pucBase64EncodedPayload, &ulBase64EncodedPayloadLength,
+                                            &pucBase64EncodedSignature, &ulBase64SignatureLength );
+    prvSwapToUrlEncodingChars( pucBase64EncodedSignature, ulBase64SignatureLength );
+
+    /* Note that we do not use mbedTLS to base64 decode values since we need the ability to assume padding characters. */
+    /* mbedTLS will stop the decoding short and we would then need to hack in the remaining characters. */
+    LogInfo( ( "---JWS Base64 Decode Header---\n" ) );
```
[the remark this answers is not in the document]
danewalton: No definitely not needed. Just useful when debugging and knowing what stage it is in the validation. I can remove.

demos/common/tests/jws_using_mbedtls_ut.c
```diff
@@ -0,0 +1,78 @@
...
+#include <string.h>
+
+/* Kernel includes. */
+#include "FreeRTOS.h"
+#include "task.h"
+
+#include "mbedtls/ctr_drbg.h"
+#include "mbedtls/entropy.h"
+
+#include "jws.h"
```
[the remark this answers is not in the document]
danewalton: correct. good point. This is only for adu so I will make that more clear

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+ * @param pucE The exponent used for the key.
+ * @param ulELength The length of \p pucE.
+ * @param pucBuffer The buffer used as scratch space to make the calculations. It should be at least
+ * `jwsRSA3072_SIZE` + `jwsSHA256_SIZE` in size.
+ * @param ulBufferLength The length of \p pucBuffer.
+ * @return uint32_t The result of the operation.
+ * @retval 0 if successful.
+ * @retval Non-0 if not successful.
+ */
+static uint32_t prvJWS_RS256Verify( unsigned char * pucInput,
```
jspaith, line 183: Do we mean RS256 or SHA256 - SHA256 is helper we're calling into and is what I normally hear this referred to.

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+    {
+        LogError( ( "mbedtls_rsa_check_pubkey res: %i\n", lMbedTLSResult ) );
+        mbedtls_rsa_free( &ctx );
+        return lMbedTLSResult;
+    }
+
+    LogInfo( ( "--- Decrypting ---\n" ) );
+
+    /* RSA */
+    lMbedTLSResult = mbedtls_rsa_pkcs1_decrypt( &ctx, NULL, NULL, MBEDTLS_RSA_PUBLIC, &ulDecryptedLength, pucSignature, pucBuffer, jwsRSA3072_SIZE );
```
jspaith, line 250: My read of `mbedtls_rsa_pkcs1_decrypt` docs [here](https://github.com/Mbed-TLS/mbedtls/blob/ca3c6a5698ef728bcede4a3db2bee964ee489a45/include/mbedtls/rsa.h#L696) is that the 2nd param, `f_rng` is mandatory.  Are MBED docs off or if not, what's consequence of this bein NULL?

 (I'm all for it being NULL FWIW to keep our life easier but I don't understand impact.)

jspaith (commented):
Additional round of comments, still mostly in the core .c file.

demos/projects/PC/linux/CMakeLists.txt
```diff
@@ -45,11 +45,38 @@ target_link_libraries(${PROJECT_NAME} PRIVATE
 add_map_file(${PROJECT_NAME} ${PROJECT_NAME}.map)
 
 # ADU demo files and dependencies
-add_executable(${PROJECT_NAME}-adu
-  main.c
-  ${CMAKE_CURRENT_LIST_DIR}/port/azure_iot_flash_platform.c
-)
-target_link_libraries(${PROJECT_NAME}-adu PRIVATE
+# add_executable(${PROJECT_NAME}-adu
```
jspaith, line 48: Nit - is there a reason this was added but commented out right away?

demos/projects/PC/linux/tests/aux_needed_functions.c
```diff
@@ -0,0 +1,256 @@
...
+
+    ( ( void ) data );
+
+    *olen = 0;
+
+    file = fopen( "/dev/urandom", "rb" );
+
+    if( file == NULL )
+    {
+        return( MBEDTLS_ERR_ENTROPY_SOURCE_FAILED );
```
jspaith, line 174: nit - (assuming we keep this logic in 1st place, see my mbedtls_platform_entropy_poll comments above) recommend appropriate logging (along with `errno`) if this is hit, along with any sort of issues with `fread` failing below.

demos/projects/PC/linux/tests/aux_needed_functions.c
```diff
@@ -0,0 +1,256 @@
...
+/*-----------------------------------------------------------*/
+
+uint64_t ullGetUnixTime( void )
+{
+    return ( uint64_t ) time( NULL );
+}
+/*-----------------------------------------------------------*/
+
+/* Needed for compilation */
+int mbedtls_platform_entropy_poll( void * data,
```
jspaith, line 158: Is this for a unit test?  This feels like something we could mock out to return something more primitive  - like rand() through the bytes.

rand is a CRT concept so if we potentially wanted this UT to run on Windows or a different board it'd be fine, but the /dev/urandom is a Linux|BSD thing.

demos/projects/PC/linux/tests/aux_needed_functions.c
```diff
@@ -0,0 +1,256 @@
+/* Standard includes. */
```
jspaith, line 1: So my read of this is that this is a bunch of FreeRTOS land stuff we need to mock out to get a UT building correctly, right?

If so could we replace "aux" with "mock" or some such?  "aux" makes it sound like they're legit helpers that we need to make work, whereas this is just a compile thing.

Also recommend that we put comments at top of file indicating the intent of this file.

demos/projects/PC/linux/tests/aux_needed_functions.c
```diff
@@ -0,0 +1,256 @@
...
+static const uint8_t ucDNSServerAddress[ 4 ] = { configDNS_SERVER_ADDR0, configDNS_SERVER_ADDR1, configDNS_SERVER_ADDR2, configDNS_SERVER_ADDR3 };
+
+/* Set the following constant to pdTRUE to log using the method indicated by the
+ * name of the constant, or pdFALSE to not log using the method indicated by the
+ * name of the constant.  Options include to standard out (xLogToStdout), to a disk
+ * file (xLogToFile), and to a UDP port (xLogToUDP).  If xLogToUDP is set to pdTRUE
+ * then UDP messages are sent to the IP address configured as the UDP logging server
+ * address (see the configUDP_LOGGING_ADDR0 definitions in FreeRTOSConfig.h) and
+ * the port number set by configPRINT_PORT in FreeRTOSConfig.h. */
+const BaseType_t xLogToStdout = pdTRUE, xLogToFile = pdFALSE, xLogToUDP = pdFALSE;
```
jspaith, line 48: Do we need all the comments above, since if this is indeed a mocked thing none of it actually applies, correct?

Do we need `xLogToStdOut` and other variables generally, either.  Only place I see them referenced is in the parameter list of `vLoggingInit` where they're going to be scoped-out by the local vars in any case.

As much as possible, any copy/paste comments that are N/A for this mocking thing (e.g. "The default IP and MAC address used by the demo" and maybe others) should be removed.

demos/projects/PC/linux/tests/aux_needed_functions.c
```diff
@@ -0,0 +1,256 @@
...
+ * This is not a secure method of generating random numbers and production
+ * devices should use a true random number generator (TRNG).
+ */
+static void prvSRand( UBaseType_t ulSeed );
+
+/* The default IP and MAC address used by the demo.  The address configuration
+ * defined here will be used if ipconfigUSE_DHCP is 0, or if ipconfigUSE_DHCP is
+ * 1 but a DHCP server could not be contacted.  See the online documentation for
+ * more information. */
+static const uint8_t ucIPAddress[ 4 ] = { configIP_ADDR0, configIP_ADDR1, configIP_ADDR2, configIP_ADDR3 };
```
jspaith, line 36: If this is just for compiling, do we need this?  It's not referenced and it's a static variable.  General question for other static only stuff that's not called inside this file.

demos/projects/PC/linux/tests/aux_needed_functions.c
jspaith: I also like the `/* Needed for compilation */` comment above one of the functions and would consider for all of them.

If I'm in a source editor and trying to lookup implementation to say `SystemWhateverFoo`, I'll jump to middle of file | the function itself and having the brief preamble above it is helpful so I know to look at next one.

libs/azure-iot-middleware-freertos/source/include/azure_iot_adu_client.h
```diff
@@ -111,6 +111,36 @@
 #define azureiotaduUPDATE_NAME_SIZE                 24
 #define azureiotaduUPDATE_VERSION_SIZE              10
 
+/* ADU.200702.R: root production key from 7/2/2020  */
+static const unsigned char AzureIoTADURootKeyId[] = "ADU.200702.R";
+static const unsigned char AzureIoTADURootKeyN[] = { 0x00, 0xd5, 0x42, 0x2e, 0xaf, 0x11, 0x54, 0xa3, 0x50, 0x65, 0x87, 0xa2, 0x4d, 0x5b, 0xba,
```
jspaith, line 116: Is this OK declaring a (large) static in a .h like this?  I worry specifically that if multiple .c files #include and reference this variable, then there'll be a separate copy of each of these variables *per .C|.obj file.

In Visual Studio land, there's a mechanism of [`__declspec(selectany)`](https://docs.microsoft.com/en-us/cpp/cpp/selectany?view=msvc-170) to tell the linker to be smarter about this and just put this in one compilation unit.   I vaguely recall a discussion with our Azure SDK friends about this a while ago but I can't recall how other compilers - gcc/clang/some of the embedded specific ones - will do right thing here.  Maybe a discussion with Azure SDK unless you know?

demos/common/tests/jws_using_mbedtls_ut.c
```diff
@@ -0,0 +1,78 @@
...
+    mbedtls_threading_set_alt( mbedtls_platform_mutex_init,
+                               mbedtls_platform_mutex_free,
+                               mbedtls_platform_mutex_lock,
+                               mbedtls_platform_mutex_unlock );
+
+    /* Initialize contexts for random number generation. */
+    mbedtls_entropy_init( pxEntropyContext );
+    mbedtls_ctr_drbg_init( pxCtrDrgbContext );
+
+    /* Add a strong entropy source. At least one is required. */
```
jspaith, line 38: nit - Does the UT really have/need a strong entropy source?

demos/common/tests/jws_using_mbedtls_ut.c
```diff
@@ -0,0 +1,78 @@
...
+#include "mbedtls/entropy.h"
+
+#include "jws.h"
+#include "demo_config.h"
+#include "FreeRTOSConfig.h"
+
+static mbedtls_entropy_context xEntropyContext;
+static mbedtls_ctr_drbg_context xCtrDrgbContext;
+
+static const char ucManifest[] = "{\"manifestVersion\":\"4\",\"updateId\":{\"provider\":\"ESPRESSIF\",\"name\":\"ESP32-Azure-IoT-Kit\",\"version\":\"1.1\"},\"compatibility\":[{\"deviceManufacturer\":\"ESPRESSIF\",\"deviceModel\":\"ESP32-Azure-IoT-Kit\"}],\"instructions\":{\"steps\":[{\"handler\":\"microsoft/swupdate:1\",\"files\":[\"fc3558477982e3235\"],\"handlerProperties\":{\"installedCriteria\":\"1.0\"}}]},\"files\":{\"fc3558477982e3235\":{\"fileName\":\"azure_iot_freertos_esp32.bin\",\"sizeInBytes\":866128,\"hashes\":{\"sha256\":\"exKJAqfEo69Ok6C6SWy9+Hhp051JbRsXsnMjGSbbJ6o=\"}}},\"createdDateTime\":\"2022-06-03T00:20:33.8421122Z\"}";
```
jspaith, line 20: Per offline discussion, we need more test cases here.  An obvious one is the SHA not matching but even things for instance like the token `sjwk` being malformatted in various ways in terms of how many dots are in it on the many cases of that.

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+ * @return uint32_t The result of the operation.
+ * @retval 0 if successful.
+ * @retval Non-0 if not successful.
+ */
+static uint32_t prvJWS_SHA256Calculate( const unsigned char * pucInput,
+                                        uint32_t ulInputLength,
+                                        unsigned char * pucOutput )
+{
+    mbedtls_md_context_t ctx;
+    mbedtls_md_type_t md_type = MBEDTLS_MD_SHA256;
```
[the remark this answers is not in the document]
danewalton: this is an mbedtls API macro

demos/common/utilities/jws_using_mbedtls.c
danewalton: This is purely based on what I observed with the manifest that was incoming. This isn't a spec dictated thing, but comes from how ADU composes the nesting doll situation.

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+        }
+        else
+        {
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( &xJSONReader ) );
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_SkipChildren( &xJSONReader ) );
+            xResult = AzureIoTJSONReader_NextToken( &xJSONReader );
+        }
+    }
+
+    if( ( xResult != eAzureIoTSuccess ) && ( xResult != eAzureIoTErrorJSONReaderDone ) )
```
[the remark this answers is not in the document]
danewalton: I will change to check the token type

demos/projects/PC/linux/tests/aux_needed_functions.c
danewalton: removed everything that isn't needed.

demos/projects/PC/linux/tests/aux_needed_functions.c
danewalton: done

demos/common/utilities/jws_using_mbedtls.c
danewalton: This is RS256 since it breaks down an RS256 payload to verify (it does use SHA256 but isn't the core of the verification)

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+                                    unsigned char * pucBuffer,
+                                    uint32_t ulBufferLength )
+{
+    AzureIoTResult_t xResult;
+    int32_t lMbedTLSResult;
+
+    if( ulBufferLength < jwsRSA3072_SIZE + jwsSHA256_SIZE )
+    {
+        LogInfo( ( "Buffer Not Large Enough\n" ) );
+        return 1;
```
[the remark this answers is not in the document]
danewalton: I don't know if we should be returning mbed errors ourselves? Is that something generally done?

demos/common/tests/jws_using_mbedtls_ut.c
danewalton: Yes for the mbedtls intialization.

demos/common/utilities/jws_using_mbedtls.c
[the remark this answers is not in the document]
jspaith: Having errors come from multiple sources - like sometimes MBEDtls, sometimes 1, sometimes some other sys thing - and here specifically at the `JWS_ManifestAuthenticate` level though the callees matter too - can be confusing.

I think maybe we should just translate - as best we can since there certainly won't be 1-1 mappings - with the `AzureIoTResult_t` code.  This would flow up through system cleaner.  We'd log the mbed error, and probably a lot of times may have to translate into some generic failure. 

We could also I guess extend out `AzureIoTResult_t` with a few cases if interesting enough and we thought caller may need it, though not with 1000's of possible errors.

libs/azure-iot-middleware-freertos/source/include/azure_iot_adu_client.h
jspaith: Yes that seems cleanest to solve this since there's no universal declspec for what MSVC is doing.

demos/common/tests/jws_using_mbedtls_ut.c
jspaith: Do we really need mbedtls?  Could we just mock it out, since ideally a UT shouldn't take any dependencies.

danewalton: I'm not so sure this is a "true" UT. This is to make it easier to verify that this functionality works. So in a way it's more of an e2e test (I can rename if necessary). This could be used in the future if someone were to need to use something like wolfssl and verify that their port worked. This is put in the same area as socket and tls porting code, which aren't tested other than by use in the samples.

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+
+    mbedtls_rsa_free( &ctx );
+
+    LogInfo( ( "---- Calculating SHA256 over input ----\n" ) );
+    xResult = prvJWS_SHA256Calculate( pucInput, ulInputLength,
+                                      pucShaBuffer );
+
+    LogInfo( ( "--- Checking if SHA256 of header+payload matches decrypted SHA256 ---\n" ) );
+
+    /* TODO: remove this once we have a valid PKCS7 parser. */
```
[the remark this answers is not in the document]
danewalton: Created a card to run through them all.

demos/common/tests/jws_using_mbedtls_ut.c
jspaith: Ah - makes sense.  I wouldn't name this an _e2e since it's not touching the backend server for anything.

At least in the C SDK I've called something like this an "_int" for integration test.  

You can call it what you want as long as it's not _ut or _e2e :).

demos/common/tests/jws_using_mbedtls_ut.c
danewalton: added a couple more negative tests

demos/common/utilities/jws_using_mbedtls.c
```diff
@@ -0,0 +1,634 @@
...
+    int doTheyMatch = memcmp( pucBuffer + jwsPKCS7_PAYLOAD_OFFSET, pucShaBuffer, jwsSHA256_SIZE );
+
+    if( doTheyMatch == 0 )
+    {
+        LogInfo( ( "SHA of JWK matches\n" ) );
+        xResult = 0;
+    }
+    else
+    {
+        LogError( ( "SHA of JWK does NOT match\n" ) );
```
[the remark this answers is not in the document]
danewalton: I mightttt agree but the logging here for that takes a bit of shuffling things around. Either some for loop printing it hex by hex or base64 encoding it. I guess if people really need to dive in, they are free to edit this file (since it's not API source code)

demos/common/utilities/jws_using_mbedtls.c
danewalton: Yea it's a tradeoff here. Doing it like this makes the flow easier to follow (you're just moving down the layers of the JWS). Breaking it up and doing only what's necessary means kicking up and down the "stack" of the JWS. I can try and rework to reduce memory usage.

jspaith: In general I'm all for flow in code (and I try to code that way myself if not always successfully which is why you guys keep me honest:)).  But I think/hope here that once this long function gets broken up into smaller helpers, then the flow will me more natural to read AND we can get memory savings at same time.  Though given it's 100's of bytes we can save, I think saving that has to be guiding principle if there's no clever way to flow+save mem at same time.

danewalton: I reduced the usage in the latest iteration by moving some of the parsing (`3875->3358` for `~500` bytes). It's possible that more could be done by parsing things out and memcpy'ing relevant values to other parts of the scratch space. But that would probably require a greater amount of fine-tooth combing. I'm willing to go down that rabbit hole but it might be another iteration.

demos/common/utilities/sample_adu_jws_mbedtls.c
```diff
@@ -0,0 +1,796 @@
...
+        LogError( ( "[JWS] Parse JSK JSON Payload Error: 0x%08x", xResult ) );
+        return xResult;
+    }
+
+    *pxSHA = pxPayload->_internal.xCoreReader.token.slice;
+
+    return 0;
+}
+
+uint32_t JWS_ManifestAuthenticate( const uint8_t * pucManifest,
```
jspaith, line 427: This is still really long - even if we have just simple helpers like as one example, `prvVerifyShaMatch` that could do from line 738 till end of function - that'd make this easier to follow.

jspaith, line 427: I think there was some other comment thread on this (GitHub UX isn't helping me find it), but I don't think we should mix and match `AzureIoTResult_t` errors along with raw MBED TLS results.  I'd rather have an MBED error translated into a less meaningful but at least consistent `eAzureIoTErrorFailed`.

The worst case would be if MBed ever returned an error of value say `1` which would be ambigious with our enum error set.

demos/common/utilities/sample_adu_jws_mbedtls.c
```diff
@@ -0,0 +1,796 @@
...
+        {
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( pxPayload ) );
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_SkipChildren( pxPayload ) );
+            xResult = AzureIoTJSONReader_NextToken( pxPayload );
+        }
+    }
+
+    azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_TokenType( pxPayload, &xJSONTokenType ) );
+
+    if( ( xResult != eAzureIoTSuccess ) && ( xJSONTokenType == eAzureIoTJSONTokenSTRING ) )
```
jspaith, line 283: I don't understand this line.  (some further comments in a response below...)

First, if `xResult` isn't SUCCESS then you cannot trust the value of `xJSONTokenType` because the callee `AzureIoTJSONReader_TokenType` is not guaranteed to have set it.  So you'd be checking against uninitialized value of it on the stack.  I'm not looking up in spec but isn't this also supposed to be a string?

Naively (see part about me not reading spec) is this

```c
if( ( xResult != eAzureIoTSuccess ) || ( xJSONTokenType != eAzureIoTJSONTokenSTRING ) )
```
==

More feedback - we're checking `azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_TokenType( pxPayload, &xJSONTokenType ) );` before we're checking the `xResult`.  Since we don't know the state of the pxPayload since we could be in an error state, I think it's safer to just check the `xResult` on its own as a separate line 1st thing out of getting out of this loop.

demos/common/utilities/sample_adu_jws_mbedtls.c
```diff
@@ -0,0 +1,796 @@
...
+    /* The "sjwk" is the signed signing public key */
+    AzureIoTJSONReader_Init( &xJSONReader, ucJWSHeader, outJWSHeaderLength );
+
+    if( prvFindJWSValue( &xJSONReader, &xJWKManifestSpan ) != 0 )
+    {
+        LogError( ( "Error finding sjwk value in payload" ) );
+        return eAzureIoTErrorFailed;
+    }
+
+    /*------------------- Base64 Decode the JWK Payload ------------------------*/
```
jspaith, line 536: This is splitting / parsing the payload but not Base64 decoding it on line right below.

demos/common/utilities/sample_adu_jws_mbedtls.c
```diff
@@ -0,0 +1,796 @@
...
+    /*Property Name */
+    azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( pxPayload ) );
+
+    while( xResult == eAzureIoTSuccess )
+    {
+        if( AzureIoTJSONReader_TokenIsTextEqual( pxPayload, ( const uint8_t * ) jws_kid_json_value, sizeof( jws_kid_json_value ) - 1 ) )
+        {
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( pxPayload ) );
+            *pxKIDSpan = pxPayload->_internal.xCoreReader.token.slice;
+
```
jspaith, line 311: (Mega) nit - extra line here.

demos/common/utilities/sample_adu_jws_mbedtls.c
jspaith: Similiar questions around `prvFindSHA`.

demos/common/utilities/sample_adu_jws_mbedtls.c
```diff
@@ -0,0 +1,796 @@
...
+                                    uint32_t ulBufferLength )
+{
+    AzureIoTResult_t xResult;
+    int32_t lMbedTLSResult;
+    uint8_t * pucShaBuffer;
+    size_t ulDecryptedLength;
+    mbedtls_rsa_context ctx;
+    int shaMatchResult;
+
+    if( ulBufferLength < jwsSHA_CALCULATION_SCRATCH_SIZE )
```
jspaith, line 185: Do we need this check?  This is a static function where we should trust caller in general, and in this case specifically `ulBufferLength` the caller has already hard-coded to `jwsSHA_CALCULATION_SCRATCH_SIZE`.

danewalton: not really I guess. Was just to double check. We can just tell people that unexpected behavior will occur if you don't use that macro

demos/common/utilities/sample_adu_jws_mbedtls.c
danewalton: Changed it to return an `AzureIoTResult_t` but the mbedtls error is printed.

demos/common/utilities/sample_adu_jws_mbedtls.c
```diff
@@ -0,0 +1,796 @@
...
+        }
+        else
+        {
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_NextToken( pxPayload ) );
+            azureiotresultRETURN_IF_FAILED( AzureIoTJSONReader_SkipChildren( pxPayload ) );
+            xResult = AzureIoTJSONReader_NextToken( pxPayload );
+        }
+    }
+
+    if( ( xResult != eAzureIoTSuccess ) && ( xResult != eAzureIoTErrorJSONReaderDone ) )
```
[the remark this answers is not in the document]
danewalton: That's theoretically done at the top with the `while()` portion;

```c
    while( xResult == eAzureIoTSuccess &&
           ( az_span_size( *pxBase64EncodedNSpan ) == 0 ||
             az_span_size( *pxBase64EncodedESpan ) == 0 ||
             az_span_size( *pxAlgSpan ) == 0 ) )
```

demos/common/utilities/sample_adu_jws_mbedtls.c
```diff
@@ -0,0 +1,886 @@
...
+
+    /*------------------- Parse root key id ------------------------*/
+
+    prvValidateRootKey( &xManifestContext );
+
+    /*------------------- Parse necessary pieces for signing key ------------------------*/
+
+    AzureIoTJSONReader_Init( &xJSONReader, xManifestContext.ucJWKPayload, xManifestContext.outJWKPayloadLength );
+
+    if( prvFindKeyParts( &xJSONReader, &xManifestContext.xBase64EncodedNSpan, &xManifestContext.xBase64EncodedESpan, &xManifestContext.xAlgSpan ) != eAzureIoTSuccess )
```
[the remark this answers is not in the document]
danewalton: Added a check to make sure it's correct.

demos/common/utilities/sample_adu_jws_mbedtls.c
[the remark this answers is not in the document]
danewalton: Gotcha gotcha. I added a check for all the sizes to make sure they're valid.

jspaith (approved):

PR #198, opened 2022-07-07
Compile ADU for Linux

danewalton:
- Adds compile for Linux.
- Adds JWS verification for main sample.


danewalton:
Holding on the PR while I smooth it out.

Issue #210, opened 2022-07-15
[ADU] Agent sample should use install criteria version instead of update id for evaluating update 


ewertons:
> Please provide us with the following information:
> ---------------------------------------------------------------
This function is incorrect:
https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/ota/demos/sample_azure_iot_adu/sample_azure_iot_pnp_simulated_data.c#L239

It should use the criteria from the steps to decide to install or not:

From a sample update manifest:

```json
{
    "steps": [{
            "type": "inline",
            "handler": "microsoft/swupdate:1",
            "files": [
                "azure_iot_freertos_esp32-v1.1.bin"
            ],
            "handlerProperties": {
                "installedCriteria": "00.06.06.42"
            }
        }
    ]
},
```

### This issue is for a: (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
>

### Any log messages given by the failure
>

### Expected/desired behavior
>

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)

### Versions
>

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

PR #211, opened 2022-07-15
Fix for ESP32 ADU sample to use installedCriteria for version comparison

ewertons:
## Purpose
Fix for github issue
https://github.com/Azure-Samples/iot-middleware-freertos-samples/issues/210

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Just run the ESP32 ADU sample as usual.

## What to Check
Verify the update happens as expected from version 1.0 to 1.1, and the update request gets rejected after the device boots up with the new image version (1.1).

### DIFF
--- a/demos/projects/ESPRESSIF/adu/config/demo_config.h
+++ b/demos/projects/ESPRESSIF/adu/config/demo_config.h
@@ -492,6 +492,6 @@ static unsigned char root_cert_array[] = {
 #define democonfigADU_DEVICE_MODEL        "ESP32-Azure-IoT-Kit"
 #define democonfigADU_UPDATE_PROVIDER     "ESPRESSIF"
 #define democonfigADU_UPDATE_NAME         "ESP32-Azure-IoT-Kit"
-#define democonfigADU_UPDATE_VERSION      "1.0"
+#define democonfigADU_UPDATE_VERSION      "1.1"
 
 #endif /* DEMO_CONFIG_H */
--- a/demos/sample_azure_iot_adu/sample_azure_iot_pnp_simulated_data.c
+++ b/demos/sample_azure_iot_adu/sample_azure_iot_pnp_simulated_data.c
@@ -235,24 +235,26 @@ static uint32_t prvGetNewMaxTemp( double xUpdatedTemperature,
 }
 /*-----------------------------------------------------------*/
 
-
-static bool prvIsUpdateAlreadyInstalled( const AzureIoTADUUpdateRequest_t * pxAduUpdateRequest )
+/**
+ * @brief Verifies if the current image version matches the "installedCriteria" version in the
+ *        installation step of the ADU Update Manifest.
+ * 
+ * @param pxAduUpdateRequest Parsed update request, with the ADU update manifest.
+ * @return true If the current image version matches the installedCriteria. 
+ * @return false If the current image version does not match the installedCriteria. 
+ */
+static bool prvDoesInstallationCriteriaMatchCurrentVersion( const AzureIoTADUUpdateRequest_t * pxAduUpdateRequest )
 {
-    if( ( pxAduUpdateRequest->xUpdateManifest.xUpdateId.ulNameLength ==
-          xADUDeviceProperties.xCurrentUpdateId.ulNameLength ) &&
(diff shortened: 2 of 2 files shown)


demos/projects/ESPRESSIF/adu/config/demo_config.h
```diff
@@ -492,6 +492,6 @@ static unsigned char root_cert_array[] = {
 #define democonfigADU_DEVICE_MODEL        "ESP32-Azure-IoT-Kit"
 #define democonfigADU_UPDATE_PROVIDER     "ESPRESSIF"
 #define democonfigADU_UPDATE_NAME         "ESP32-Azure-IoT-Kit"
-#define democonfigADU_UPDATE_VERSION      "1.0"
+#define democonfigADU_UPDATE_VERSION      "1.1"
```
danewalton, line 495: should we leave this as `1.0`

demos/sample_azure_iot_adu/sample_azure_iot_pnp_simulated_data.c
```diff
@@ -277,9 +279,9 @@ static bool prvIsUpdateAlreadyInstalled( const AzureIoTADUUpdateRequest_t * pxAd
  */
 static AzureIoTADURequestDecision_t prvUserDecideShouldStartUpdate( AzureIoTADUUpdateRequest_t * pxAduUpdateRequest )
 {
-    if( prvIsUpdateAlreadyInstalled( pxAduUpdateRequest ) )
+    if( !prvDoesInstalledCriteriaMatchCurrentVersion( pxAduUpdateRequest ) )
```
danewalton, line 282: If it does match, the update manifest was applied and the device rebooted correct?
```suggestion
    if( prvDoesInstalledCriteriaMatchCurrentVersion( pxAduUpdateRequest ) )
```

demos/sample_azure_iot_adu/sample_azure_iot_pnp_simulated_data.c
```diff
@@ -277,9 +279,9 @@ static bool prvIsUpdateAlreadyInstalled( const AzureIoTADUUpdateRequest_t * pxAd
...
     {
-        LogInfo( ( "[ADU] Rejecting update request (current version is up-to-date)" ) );
+        LogInfo( ( "[ADU] Rejecting update request (installed criteria does not match current version)" ) );
```
danewalton, line 284:
```suggestion
        LogInfo( ( "[ADU] Rejecting update request (installed criteria matches current version)" ) );
```

demos/projects/ESPRESSIF/adu/config/demo_config.h
ewertons: Fixed.

demos/sample_azure_iot_adu/sample_azure_iot_pnp_simulated_data.c
ewertons: If the device rebooted, now it has the new version (1.1), so the "installedCriteria" will not match anymore. The name from ADU may be misleading, I believe.

danewalton: https://docs.microsoft.com/en-us/previous-versions/azure/iot-hub-device-update/import-schema#import-manifest-schema

From there ^

> String interpreted by the agent to determine whether the update was applied successfully

I'm assuming that is a string that is of importance for the device to compare in its own way. If that "installedCriteria" matches, that means it updated successfully.

danewalton: I think if it's the way you're thinking, it would be called `installCriteria`? In that case if it matched, it would mean apply the update.

ewertons:
After a meeting with ADU engineers, we got this summary:
we will modify our sample code to handle the update manifest as such:
- Sample gets the version of the current device image (e.g, 1.0)
- Sample checks the step installedCriteria (e.g., 1.1)
- If the data is different in these two, consume the update request. Otherwise, ignore.

PR #219, opened 2022-07-19
Expose ADU compatibility properties as options to user

ewertons:
## Purpose

* ...

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/libs/azure-iot-middleware-freertos/libraries/azure-sdk-for-c/sdk/src/azure/iot/az_iot_adu_client.c
+++ b/libs/azure-iot-middleware-freertos/libraries/azure-sdk-for-c/sdk/src/azure/iot/az_iot_adu_client.c
@@ -14,9 +14,6 @@
 /* Define the ADU agent interface ID.  */
 #define AZ_IOT_ADU_CLIENT_AGENT_INTERFACE_ID "dtmi:azure:iot:deviceUpdate;1"
 
-/* Define the compatibility.  */
-#define AZ_IOT_ADU_CLIENT_AGENT_COMPATIBILITY "manufacturer,model"
-
 /* Define the ADU agent property name "agent" and sub property names.  */
 #define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_AGENT "agent"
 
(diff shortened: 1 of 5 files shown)


libs/azure-iot-middleware-freertos/libraries/azure-sdk-for-c/sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -240,12 +237,21 @@ AZ_NODISCARD az_result az_iot_adu_client_get_agent_state_payload(
...
       ref_json_writer,
       AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_COMPAT_PROPERTY_NAMES)));
-  _az_RETURN_IF_FAILED(az_json_writer_append_string(
-      ref_json_writer, AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_COMPATIBILITY)));
+  if (az_span_is_content_equal(
+          device_properties->compatibility_properties, AZ_SPAN_EMPTY))
+  {
+    _az_RETURN_IF_FAILED(az_json_writer_append_string(
+        ref_json_writer, AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_DEFAULT_COMPATIBILITY_PROPERTIES)));
+  }
```
danewalton, line 249: for this here, should we allow it to be empty? The user should either have their own or use the default. But in either case we want them to make the selection. Kind of like the sample from above in the middleware. We show that people should use the default "by default".

libs/azure-iot-middleware-freertos/source/include/azure_iot_adu_client.h
```diff
@@ -150,6 +150,8 @@ typedef struct AzureIoTADUClientDeviceProperties
     uint32_t ulModelLength;
 
     AzureIoTADUDeviceCustomProperties_t * pxCustomProperties;
+    const uint8_t * ucCompatibilityProperties;
+    uint32_t ulCompatibilityPropertiesLength;
```
danewalton, line 154: should these live here? Since they're technically part of separate fields? 

image

Maybe have it as the first candidate for a field in our `options` struct? It being in the options also allows us to set it to the default for the case that the user doesn't set it?

ewertons: Done.

libs/azure-iot-middleware-freertos/libraries/azure-sdk-for-c/sdk/src/azure/iot/az_iot_adu_client.c
ewertons: Done.

danewalton (approved):

libs/azure-iot-middleware-freertos/source/include/azure_iot_adu_client.h
```diff
@@ -285,7 +285,8 @@ typedef struct AzureIoTADUUpdateRequest
  */
 typedef struct AzureIoTADUClientOptions
 {
-    void * xUnused;
+    const uint8_t * ucCompatibilityProperties;
```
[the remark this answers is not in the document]
ewertons: Addressed. Thanks.

PR #221, opened 2022-07-19
remove stale log

danewalton:
The response is handled inside the `vHandleWritableProperties()` function so no response is needed from this function.

https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/c7ff54f4ae986f447dd888072f6e658c6c544917/demos/projects/ESPRESSIF/adu/main/azure_iot_freertos_esp32_pnp_model.c#L276-L289

### DIFF
--- a/demos/sample_azure_iot_adu/sample_azure_iot_adu.c
+++ b/demos/sample_azure_iot_adu/sample_azure_iot_adu.c
@@ -291,19 +291,6 @@ static void prvDispatchPropertiesUpdate( AzureIoTHubClientPropertiesResponse_t *
                                ucReportedPropertiesUpdate,
                                sizeof( ucReportedPropertiesUpdate ),
                                &ulReportedPropertiesUpdateLength );
-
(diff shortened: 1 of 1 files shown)


ewertons (commented):
But what about any non-ADU properties the user might receive on their device? I'm ok removing the log, but not the logic to send a response. What do you think?

danewalton:
> But what about any non-ADU properties the user might receive on their device? I'm ok removing the log, but not the logic to send a response. What do you think?

My understanding is that the DTMI which we have in this dictates all the properties which the device may receive. As in the device shouldn't accept properties which aren't in the DTMI.

We have this line which just logs things it doesn't recognize but doesn't create a response payload:

```c
        else
        {
            LogInfo( ( "Component not ADU: %.*s", ulComponentNameLength, pucComponentName ) );
            prvSkipPropertyAndValue( &xJsonReader );
        }
```

ewertons (approved):
Let's remove the non-ADU property for now, and add back when we play more with pnp composition.

PR #237, opened 2022-09-01
add cancel process manifest

danewalton:


danewalton:
Please hold while I look for a different solution

danewalton:
## Cancel
image

## Reject
image

danewalton:
image

Successfully cancels mid download.

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -835,17 +863,50 @@ static void prvAzureDemoTask( void * pvParameters )
...
+                                                                &xAzureIoTHubClient,
+                                                                &xADUDeviceProperties,
+                                                                &xAzureIoTAduUpdateRequest,
+                                                                eAzureIoTADUAgentStateIdle,
+                                                                NULL,
+                                                                ucScratchBuffer,
+                                                                sizeof( ucScratchBuffer ),
+                                                                NULL );
+
+                    xProcessUpdateRequest = false;
```
ewertons, line 880: To make sure we are playing safe, should xImage be reset here? 
On line 510 we already check `xImage.ulCurrentOffset < xImage.ulImageFileSize`, so if `ulImageFileSize` is reset to zero, it's an extra layer of safety.

danewalton: https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/88b8c8412f18c1e34544539c2b569a89334bda1d/demos/sample_azure_iot_adu/sample_azure_iot_adu.c#L455

I think this call at the beginning of the `prvDownloadUpdateImageIntoFlash()` would be sufficient? If a cancellation gets invoked in the middle of the download, it kicks out of `prvDownloadUpdateImageIntoFlash()` and then next time around, it would get re-init'd.

danewalton:
There was a worry about whether this code would download from a different URL in the middle of a download (should a new update come in).

Couple resolutions for that question:

1. First, there should be a cancel command which comes in before the update is changed. It would go `Download 1.1` -> `Cancel` -> `Device pings that it's idle` -> `Device receives 1.2 and downloads`.
2. We parse the URL into its host name and path. So even if it swaps from 1.1 to 1.2, we still use the original URL and there won't be cross contamination. If the URL does change and we kick out of that function, upon subsequent invocation, the HTTP state context will be reset and we're okay.

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -835,17 +863,50 @@ static void prvAzureDemoTask( void * pvParameters )
...
+                }
+                else
+                {
+                    xResult = prvDownloadUpdateImageIntoFlash( sampleazureiotADU_DOWNLOAD_TIMEOUT );
+                    configASSERT( xResult == eAzureIoTSuccess );
+
+                    if( xAzureIoTAduUpdateRequest.xWorkflow.xAction == eAzureIoTADUActionApplyDownload )
```
vaavva, line 887: Is there a scenario where xProcessUpdateRequest is true, xAzureIoTAduUpdateRequest.xWorkflow.xAction != eAzureIoTADUActionCancel and xAzureIoTAduUpdateRequest.xWorkflow.xAction isn't eAzureIoTADUActionApplyDownload?

danewalton: No there's only two values for that enum. That being said, I don't think it's a bad idea to have an explicit check for each and then an else which logs that an unknown value was set. I'll update....good question.

danewalton: Your question also brings up another scenario that I think we need to account for here. We should check for `xProcessUpdateRequest` again in case an invalid ADU payload came in (with some kind of weird `NULL`s or something. I'll add it

vaavva: Agree - it's good future proofing in case a third value comes along

vaavva: > Your question also brings up another scenario that I think we need to account for here. We should check for `xProcessUpdateRequest` again in case an invalid ADU payload came in (with some kind of weird `NULL`s or something. I'll add it

This would be in the scenario that during one of the checks during the timeouts in prvDownloadUpdateImageIntoFlash we get a cancel request? Would that scenario bump us out of this loop though because the assert on 885 `configASSERT( xResult == eAzureIoTSuccess );` wouldn't be true?

vaavva: I think it also might be good to add a comment calling out why we have the last if statement since the code looks like we check `xProcessUpdateRequest && ( xAzureIoTAduUpdateRequest.xWorkflow.xAction == eAzureIoTADUActionApplyDownload )` right next to the other if statements checking these values

danewalton: Yes in the case that the deployment was cancelled in the middle of the download, it breaks the download and returns success. So the assert would not get tripped. Also added extra comments on why we're doing what we're doing.

vaavva (approved):

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -122,6 +122,12 @@
  */
 #define sampleazureiotSUBSCRIBE_TIMEOUT                       ( 10 * 1000U )
 
+/**
+ * @brief Timeout for downloading the update image and calling the
+ * AzureIoTHubClient_ProcessLoop() function
```
ewertons, line 127: Is it always assumed that these timeouts are in seconds, or could we add that to the description?

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -495,8 +503,28 @@ static AzureIoTResult_t prvDownloadUpdateImageIntoFlash()
...
+
     while( xImage.ulCurrentOffset < xImage.ulImageFileSize )
     {
+        ullCurrentTime = ullGetUnixTime();
+
+        if( ullCurrentTime - ullPreviousTimeout > ullTimeoutInSec )
+        {
+            LogInfo( ( "%i second timeout. Taking a break from downloading image.", ullTimeoutInSec ) );
+            LogInfo( ( "Receiving messages from IoT Hub." ) );
+            xResult = AzureIoTHubClient_ProcessLoop( &xAzureIoTHubClient,
```
ewertons, line 516: This is the easiest way to do this, but would it be the only way?
Have you thought about completely leaving prvDownloadUpdateImageIntoFlash and having only the  AzureIoTHubClient_ProcessLoop call in the main task?

danewalton: Yea I thought about that. I think it trades some simplicity for a different kind of complexity. Either you kick out and track the HTTP state outside the function, where resetting it also has to be done outside. Or you have multiple places to call the `_ProcessLoop()`. I can have another look and think about returning from this function instead.

danewalton: Yea looking at it again (and trying it a bit), it would add a different kind of complexity. I'm not totally opposed, but I only think it's worth it if people feel passionate about it.

ewertons: I think the decision is a function of time. If we are short on time, then this design should be enough. If we have more time, then the one you described would make this sample a little closer to a model customers could reuse.
I will approve this design and let you decide on what makes more sense. If you make any changes and need more review, please let me know.

ewertons (approved):

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
danewalton: Understood. I think this works for the intended purpose, and with a subsequent update we can change if that's better 👍

PR #241, opened 2022-09-09
unify names of helper files

danewalton:
Customer was confused that they couldn't find `crypto.h` which they thought was a wolfssl include. We should probably rename it to avoid confusion.

This also updates the other helper function to make it similar.


vaavva:
Is this a standard that we use for other helper files across the samples?

vaavva:
> Maybe sample_common_adu_jws.h and sample_common_crypto.h?

Hmm, I like the idea of sample_common to differentiate it from the actual samples, but makes it clear that it's our code - I think the azure_sample prefix also works to differentiate since the sample files start with sample_azure

danewalton:
> > Maybe sample_common_adu_jws.h and sample_common_crypto.h?
> 
> Hmm, I like the idea of sample_common to differentiate it from the actual samples, but makes it clear that it's our code - I think the azure_sample prefix also works to differentiate since the sample files start with sample_azure

I'll probably go through with as is and we can always change it if people don't like it.

vaavva (approved):

PR #255, opened 2022-10-24
STM32L475 ADU sample

vaavva:
## Purpose

* Adds ADU sample for the STM32L475
* Readme to come in a follow-up PR, as this only merges previously reviewed code

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


wduraes (approved):
since code changes were previously approved, I'm approving and will wait for the merge so I can test on the device.

demos/projects/ST/b-l475e-iot01a/ADU.md
```diff
@@ -0,0 +1,12 @@
+
+Generate the update manifest using **powershell**.
+
+```powershell
+Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope Process
+Import-Module .\iot-hub-device-update\tools\AduCmdlets\AduUpdate.psm1
+$updateId = New-AduUpdateId -Provider "STMicroelectronics" -Name "STM32L475" -Version 1.6
+$compat = New-AduUpdateCompatibility -Properties @{ deviceManufacturer = 'STMicroelectronics'; deviceModel = 'STM32L475' }
+$installStep = New-AduInstallationStep -Handler 'microsoft/swupdate:1'-HandlerProperties @{ installedCriteria = '1.6' } -Files C:\repos\iot-middleware-sample-adu.1.6.bin
```
danewalton, line 9: nit: make 1.6 -> 1.1

danewalton (approved):
Before merging, please restore this: 

https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/d73de4bcf47e68503fa4afa43265b2ef8eb6ca93/.github/workflows/ci_tests_linux.yml#L22-L23

to add st to the gate. You can remove the line after it which had dropped st. You'll have to add `iot-middleware-sample-adu` to here as well : https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/d73de4bcf47e68503fa4afa43265b2ef8eb6ca93/.github/scripts/ci_tests.sh#L83

as a new check

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * The following header line is automatically written to
+ * #HTTPRequestHeaders_t.pBuffer:
+ * "User-Agent: my-platform-name\r\n"
+ *
+ * <b>Possible values:</b> Any string. <br>
+ * <b>Default value:</b> `my-platform-name`
+ */
+#ifndef HTTP_USER_AGENT_VALUE
+    #define HTTP_USER_AGENT_VALUE    "stml475-adu"
```
CIPop, line 70: Is this tracked in any way by ADU? Should it be more specific (i.e. including version, the fact it's a sample, etc)

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
```diff
@@ -0,0 +1,149 @@
...
+ * If #HTTPResponse_t.getTime is set to NULL, then this HTTP_RECV_RETRY_TIMEOUT_MS
+ * is unused. When this timeout is unused, #HTTPClient_Send will not retry the
+ * transport receive calls that return zero bytes read.
+ *
+ * <b>Possible values:</b> Any positive 32 bit integer. A small timeout value
+ * is recommended. <br>
+ * <b>Default value:</b> `10`
+ */
+#ifndef HTTP_RECV_RETRY_TIMEOUT_MS
+    #define HTTP_RECV_RETRY_TIMEOUT_MS    ( 10U )
```
CIPop, line 94: Please consider increasing the timeouts, especially for WiFi communication.

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
```diff
@@ -0,0 +1,149 @@
...
+ * If #HTTPResponse_t.getTime is set to NULL, then this HTTP_RECV_RETRY_TIMEOUT_MS
+ * is unused. When this timeout is unused, #HTTPClient_Send will not retry the
+ * transport send calls that return zero bytes sent.
+ *
+ * <b>Possible values:</b> Any positive 32 bit integer. A small timeout value
+ * is recommended. <br>
+ * <b>Default value:</b> `10`
+ */
+#ifndef HTTP_SEND_RETRY_TIMEOUT_MS
+    #define HTTP_SEND_RETRY_TIMEOUT_MS    ( 10U )
```
CIPop, line 117: Same as above.

demos/projects/ST/b-l475e-iot01a/config/demo_config.h
```diff
@@ -505,6 +505,18 @@ static unsigned char root_cert_array[] = {
...
+
+#ifdef democonfigENABLE_DPS_SAMPLE
+    #define democonfigCHUNK_DOWNLOAD_SIZE    256
+#else
+    #define democonfigCHUNK_DOWNLOAD_SIZE    1024
+#endif /* democonfigENABLE_DPS_SAMPLE */
+
+#define democonfigADU_DEVICE_MANUFACTURER    "STMicroelectronics"
+#define democonfigADU_DEVICE_MODEL           "STM32L475"
+#define democonfigADU_UPDATE_PROVIDER        "STMicroelectronics"
```
CIPop, line 518: Should these include some information about our SDK and version for telemetry purposes?

demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
```diff
@@ -0,0 +1,285 @@
+/* Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License. */
+
+#include <string.h>
+
```
CIPop, line 5: Nit: we should remove whitespace between includes.

demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
```diff
@@ -0,0 +1,285 @@
...
+#include "azure/core/az_base64.h"
+
+#include "mbedtls/md.h"
+
+/* advance addr by this amount to program the next 32 row double-word (64-bit)
+ * for fast programming
+ */
+#define azureiotflashL475_FLASH_ROW_SIZE      256
+
+#define azureiotflashL475_DOUBLE_WORD_SIZE    8
```
CIPop, line 23: Should this be defined as 2*sizeof(long) or uintptr_t ?

demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
```diff
@@ -0,0 +1,285 @@
...
+    mbedtls_md_free( &ctx );
+
+    if( memcmp( ucDecodedManifestHash, ucCalculatedHash, azureiotflashSHA_256_SIZE ) == 0 )
+    {
+        AZLogInfo( ( "SHAs match\r\n" ) );
+        xResult = eAzureIoTSuccess;
+    }
+    else
+    {
+        AZLogInfo( ( "SHAs do not match\r\n" ) );
```
CIPop, line 220: This should not be Information but error (or at least warning). We should also try to log a single time instead of many log lines.

Also, formatting should only be done by AZLogInfo and not printf - a log listener may direct these messages elsewhere.

CIPop (commented):
Left a few comments.

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
vaavva: Any suggestions on a better value?

demos/projects/ST/b-l475e-iot01a/config/demo_config.h
vaavva: This is what shows up in Azure Portal when you select device information, so I think it's probably better to be just about the device. We could definitely open this as a topic for all of the ADU samples (currently the rest of them are similar to this)

demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
vaavva: This was done to match the verification in the ESP32 sample, but let me see what I can do to modify it and then make those changes there as well

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
vaavva: I'm not sure - @danewalton, do you know the answer to this?

danewalton: to my knowledge, no they don't track it

CIPop: @wduraes - FYI, in case there's something we need to track.

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
CIPop: Maybe I misunderstood what these mean but I'd give it a try with 500/1000ms, especially since offline we were talking about some disconnects in certain cases.

demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
vaavva: I was able to change the first line to an error (I wasn't sure if the rest should be categorized more as additional info or more error lines).
I couldn't figure out a way to print the whole hash so we could use one logError for the whole bit of info - I think I'm running into this issue: https://eevblog.com/forum/microcontrollers/printf-uint64_t-with-arm-gcc/ 
> AZLogError( ( "SHAs do not match\r\nWanted: %.*x\r\nCalculated: %.*x\r\n", azureiotflashSHA_256_SIZE, *ucDecodedManifestHash, azureiotflashSHA_256_SIZE, *ucCalculatedHash ) );

I also tried to at least convert the printfs to AzureLogInfos, but that adds a newline at the end, so the hash ends up like this instead of on the same line 
> ea
fa
d1
fa
cd
ea
6a
cc
ab
33
...

If you have any suggestions on how to solve either of these issues, let me know!

demos/projects/ST/b-l475e-iot01a/config/demo_config.h
CIPop: In that case it should definitely be something like `Azure IoT SDK FreeRTOS ADU Sample` (or similar w/o spaces) to make sure there is no confusion between prod and test devices.

demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
CIPop: Ah, in that case you could consider the issue a future enhancement and track separately. 

In general, we should never log more than one line per error/info message. This helps log analysis tools that usually expect a single record per message.

CIPop (approved):
Thanks @vaavva !

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
vaavva: The issues I was running into were because my wifi hotspot is only connectable when my phone is open to the hotspot page for some reason 🙃 And the second issue was with disconnects during download, but that issue I think should be fixed with retries, not a longer timeout. So I'm not quite sure how to test whether a larger number is sufficient or not (especially since the comments for this say to keep it small)
@danewalton, did you have a way to repro the issues you were seeing?

demos/projects/ST/b-l475e-iot01a/config/demo_config.h
vaavva: I'll open a work item for this and the printing issue to solve these issues in a consistent way across the samples

demos/projects/ST/b-l475e-iot01a/config/core_http_config.h
vaavva: Discussed offline - will leave as 10U for now

PR #262, opened 2022-11-07
update sub

danewalton:


demos/sample_azure_iot_adu/sample_azure_iot_pnp_simulated_data.c
```diff
@@ -377,6 +466,8 @@ void vHandleWritableProperties( AzureIoTHubClientPropertiesResponse_t * pxMessag
                                                                  xAzureIoTAduUpdateRequest.ulUpdateManifestLength,
                                                                  xAzureIoTAduUpdateRequest.pucUpdateManifestSignature,
                                                                  xAzureIoTAduUpdateRequest.ulUpdateManifestSignatureLength,
+                                                                 &xADURootKeys[ 0 ],
+                                                                 sizeof( xADURootKeys ) / sizeof( xADURootKeys[ 0 ] ),
```
vaavva, line 470: Will each root key always be the same size, or could this math get a bit funky if they aren't exactly the same?

danewalton: Even if the keys aren't the same, this should be okay since it's getting the size of the struct which is a bunch of pointers and lengths to the actual key values.

vaavva (approved):

wduraes (approved):

PR #264, opened 2022-11-07
[ADU] Reject updates that are too large

vaavva:
## Purpose

* rejects any device updates that are larger than the bootable bank size on the device

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


demos/projects/ESPRESSIF/adu/port/azure_iot_flash_platform.c
```diff
@@ -71,6 +71,12 @@ AzureIoTResult_t AzureIoTPlatform_Init( AzureADUImage_t * const pxAduImage )
     return eAzureIoTSuccess;
 }
 
+int64_t AzureIoTPlatform_GetSingleFlashBootBankSize()
+{
+    /* size of the next ota partition to be written to */
+    return esp_ota_get_next_update_partition( esp_ota_get_running_partition() )->size;
```
danewalton, line 77: can this return `NULL`? If so might have a hard fault here.

danewalton (approved):

demos/projects/NXP/mimxrt1060/port/azure_iot_flash_platform.c
```diff
@@ -53,6 +53,15 @@ AzureIoTResult_t AzureIoTPlatform_Init( AzureADUImage_t * const pxAduImage )
     return eAzureIoTSuccess;
 }
 
+int64_t AzureIoTPlatform_GetSingleFlashBootBankSize()
+{
+    // TODO: Fill in returning of flash bank size (to make sure we don't install an image larger than that)
```
ericwol-msft, line 58: if this is the final PR, why is there a todo?

vaavva: This file is a stub for the NXP sample - all of the other functions have TODO for implementation as well

ericwol-msft (approved):

wduraes (approved):
Very important change. Thank you for doing this.

demos/projects/ESPRESSIF/adu/port/azure_iot_flash_platform.c
vaavva: Good catch! I updated it to check, and return -1 as the size if we run into issues. This will then fail the check that the bank size is larger than the update and reject it. It won't automatically try to get the update again, but the user can retry from the azure portal (or if the device restarts). Getting the device to fail the way we do in other steps (which would restart the device and try again) isn't super feasible at the accept/reject step without significant refactoring, so this seemed like a good compromise for a solution.

PR #267, opened 2022-11-09
Remove printfs from verifyImage

ewertons:
## Purpose

* ...
REmove `printf` from functions in azure_iot_flash_platform.c file**S**

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[ ] Other... Please describe:
```

## How to Test
Run the ADU samples for ESP32 and ST L475 and verify logs.

## What to Check
Verify that the following are valid:
No breaks in functionality due to logging changes.
And that the similar prints are still logged (timestamps and image size may vary):

I (88168) AZ IOT: Starting the mbedtls calculation: image size 866304
I (94258) AZ IOT: mbedtls calculation completed

## Other Information
Lint collected in your dryer's vent can be a fire hazard! [Clean them regularly](https://www.lowes.com/n/how-to/clean-a-dryer-vent)!

### DIFF
--- a/demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
+++ b/demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
@@ -215,7 +215,7 @@ AzureIoTResult_t AzureIoTPlatform_VerifyImage( AzureADUImage_t * const pxAduImag
 
         for( int i = 0; i < azureiotflashSHA_256_SIZE; ++i )
         {
-            AZLogInfo( "%x", ucDecodedManifestHash[ i ] );
+            AZLogInfo( ( "%x", ucDecodedManifestHash[ i ] ) );
         }
 
         AZLogInfo( ( "\r\n" ) );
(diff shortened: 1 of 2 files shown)


CIPop (approved):
Thanks @ewertons !

danewalton (approved):

demos/projects/ST/b-l475e-iot01a/port/azure_iot_flash_platform.c
```diff
@@ -223,18 +215,18 @@ AzureIoTResult_t AzureIoTPlatform_VerifyImage( AzureADUImage_t * const pxAduImag
 
         for( int i = 0; i < azureiotflashSHA_256_SIZE; ++i )
         {
-            printf( "%x", ucDecodedManifestHash[ i ] );
+            AZLogInfo( ( "%x", ucDecodedManifestHash[ i ] ) );
```
vaavva, line 218: Sorry, didn't get a chance to look at this last night, but this will print out the hash on a bunch of lines - not sure if this is what we want
image

ewertons: Thanks for bringing this up. It was intentional.
I didn't want to add more memory manipulation just to have the log printed nicely.
The best way would be to change AzLogInfo to have an option to print without adding timestamp and newline, then it would print the way we really would like it to. But I didn't want to make such changes to logging at this point.

PR #279, opened 2022-11-17
Build adu dps payload

danewalton:
- The payload before was using the demo config model id which is the wrong one. This generates the DPS payload with the value which is given from the SDK.

### DIFF
--- a/demos/sample_azure_iot_adu/sample_azure_iot_adu.c
+++ b/demos/sample_azure_iot_adu/sample_azure_iot_adu.c
@@ -974,15 +974,15 @@ static void prvAzureDemoTask( void * pvParameters )
             return xResult;
         }
         else if( ( xResult = AzureIoTJSONWriter_AppendPropertyName( &xWriter,
-                                                                ( uint8_t * ) sampleazureiotMODEL_ID_STR,
-                                                                sizeof( sampleazureiotMODEL_ID_STR ) - 1 ) ) != eAzureIoTSuccess )
+                                                                    ( uint8_t * ) sampleazureiotMODEL_ID_STR,
+                                                                    sizeof( sampleazureiotMODEL_ID_STR ) - 1 ) ) != eAzureIoTSuccess )
         {
             LogError( ( "Error appending property: result 0x%08x", xResult ) );
             return xResult;
         }
         else if( ( xResult = AzureIoTJSONWriter_AppendString( &xWriter,
-                                                                   AzureIoTADUModelID,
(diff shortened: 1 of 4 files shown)


demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -154,6 +154,8 @@ uint64_t ullGetUnixTime( void );
     static uint8_t ucSampleIotHubHostname[ 128 ];
     static uint8_t ucSampleIotHubDeviceId[ 128 ];
     static AzureIoTProvisioningClient_t xAzureIoTProvisioningClient;
+    static uint8_t ucProvisioningPayloadBuffer[ 128 ];
```
vaavva, line 157: Any chance we can use the ucScratchBuffer for this like in the gsg sample?

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -950,6 +952,52 @@ static void prvAzureDemoTask( void * pvParameters )
...
+        {
+            LogError( ( "Error initializing JSON writer: result 0x%08x", xResult ) );
+            return xResult;
+        }
+        else if( ( xResult = AzureIoTJSONWriter_AppendBeginObject( &xWriter ) ) != eAzureIoTSuccess )
+        {
+            LogError( ( "Error appending begin object: result 0x%08x", xResult ) );
+            return xResult;
+        }
+        else if( ( xResult = AzureIoTJSONWriter_AppendPropertyName( &xWriter,
```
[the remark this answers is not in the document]
danewalton: Ahhh yup I had it like that and broke it up trying to figure out why it was failing to append the property value. Good call.

vaavva (approved):

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -964,6 +1006,7 @@ static void prvAzureDemoTask( void * pvParameters )
         TlsTransportParams_t xTlsTransportParams = { 0 };
         AzureIoTResult_t xResult;
         AzureIoTTransportInterface_t xTransport;
+        uint32_t ulOutProvisioningPayloadLength;
```
vaavva, line 1009:
```suggestion
        int32_t ulOutProvisioningPayloadLength;
```
AzureIoTJSONWriter_GetBytesUsed returns an int32_t

PR #290, opened 2022-12-06
Add new sample for ca recovery

danewalton:
- Creates a new sample at the high level for CA recovery. It is a direct copy of `sample_azure_iot` except for the name.
- Creates a separate sample in the ESP build tree which pulls in and builds the above sample instead of `sample_azure_iot`.


danewalton:
Reopening to keep it simple

danewalton:
> our ESPRESSIF folder is a mess. it should be
> 
> ```
> iot-middleware-freertos-samples/demos/projects/ESPRESSIF
>                                                   |
>                                                   + ESP32
>                                                       |
>                                                       + adu
>                                                       |
>                                                       + aziotkit
>                                                       |
>                                                       + generic
>                                                       |
>                                                       + ca-recovery
> ```

Noted. We'll try and patch that up in another PR.

PR #294, opened 2022-12-13
Manually install ARM GCC

danewalton:


.github/scripts/install_software.sh
```diff
@@ -8,4 +8,12 @@ set -o nounset # Exit if variable not set.
...
 
 sudo apt update
-sudo apt install -y net-tools gcc-multilib g++-multilib ninja-build libpcap-dev ethtool isc-dhcp-server unifdef gcc-arm-none-eabi dos2unix
+sudo apt install -y net-tools gcc-multilib g++-multilib ninja-build libpcap-dev ethtool isc-dhcp-server unifdef dos2unix
+
+wget https://developer.arm.com/-/media/Files/downloads/gnu/11.2-2022.02/binrel/gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz
+
+sudo mkdir /usr/local/arm-gcc-none-eabi
+
+sudo tar -xvf gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz -C /usr/local/arm-gcc-none-eabi --strip-components=1
```
vaavva, line 17: Do we need to add `tar` to the install chain as well to be able to run this?

vaavva (approved):

.github/scripts/install_software.sh
danewalton: I believe the github runners automatically have it installed.

vaavva: Oh, I meant more for users since we reference using this script to get locally set up in the readme for E2Es

danewalton: good point. Let me try something. Going to break out the manual install into a different script and leave this one so that we don't have to muck with customer `/usr/local` directories any more than we have to. And I'll leave the `apt install` version of arm-gcc for those people where it works. Silo this special behavior just to the gate for now.

vaavva: Does the apt install version work locally, or if a user pulled it fresh would they run into the same issue? There are a few things like `tar` that I'm adding in for the v5 manifest pr as well (in that users need to install them, but the github runners don't), so I wonder if it could be worth pulling all of those out of the runner install or something - just wanted to throw that out there in case more scenarios of this would change the best implementation

danewalton: having both seems to cause a compiler crash, so i'm going to go with only the manually installed one.

I believe this should still be okay for users coming from scratch, although it's hard to know unless someone does it themselves.

PR #314, opened 2023-01-11
Prevent core dump on wifi disconnect by restarting sample

vaavva:
## Purpose

Definitions:
* network disconnect - when the network is turned off and the device cannot connect at all
* internet disconnect - when the network is still present, but doesn't provide an internet connection (ex. if you unplug the ethernet to your router, but the power is still on)

Problem: In the sample, a network disconnect or internet disconnect would cause a core dump in the sample because of failed asserts.

Solution: this PR checks for the failed values before the assert and instead of crashing the device, breaks out of the sample execution and starts it over, which forces the network to reconnect

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[x] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
Test both scenarios:
network disconnect - turn off wifi (either hotspot or unplug the router)
internet disconnect - turn off internet connection (either in router settings or by unplugging the ethernet cable that goes to the router from the modem)
```
Tested:
pnp sample:
* ST L475 network disconnect
* esp32 network disconnect
* esp32 internet disconnect
* aziotkit network disconnect
* linux network disconnect
* windows internet (?) disconnect - unplugged the ethernet cable and it behaved like an internet disconnect
* NXP 1060 internet (?) disconnect - unplugged the ethernet cable and it behaved like an internet disconnect

## What to Check
* No core dumps should occur, and device should reconnect once internet is back

### DIFF
--- a/demos/projects/ESPRESSIF/esp32/components/sample-azure-iot/backoff_algorithm.c
+++ b/demos/projects/ESPRESSIF/esp32/components/sample-azure-iot/backoff_algorithm.c
@@ -46,7 +46,7 @@ BackoffAlgorithmStatus_t BackoffAlgorithm_GetNextBackoff( BackoffAlgorithmContex
 
     /* If BACKOFF_ALGORITHM_RETRY_FOREVER is set to 0, try forever. */
     if( ( pRetryContext->attemptsDone < pRetryContext->maxRetryAttempts ) ||
-        ( BACKOFF_ALGORITHM_RETRY_FOREVER == 0 ) )
+        ( pRetryContext->maxRetryAttempts == BACKOFF_ALGORITHM_RETRY_FOREVER ) )
     {
         /* The next backoff value is a random value between 0 and the maximum jitter value
          * for the retry attempt. */
--- a/demos/projects/ESPRESSIF/esp32/components/sample-azure-iot/backoff_algorithm.c
+++ b/demos/projects/ESPRESSIF/esp32/components/sample-azure-iot/backoff_algorithm.c
@@ -46,7 +46,7 @@ BackoffAlgorithmStatus_t BackoffAlgorithm_GetNextBackoff( BackoffAlgorithmContex
 
     /* If BACKOFF_ALGORITHM_RETRY_FOREVER is set to 0, try forever. */
     if( ( pRetryContext->attemptsDone < pRetryContext->maxRetryAttempts ) ||
-        ( pRetryContext->maxRetryAttempts == BACKOFF_ALGORITHM_RETRY_FOREVER ) )
+        ( BACKOFF_ALGORITHM_RETRY_FOREVER == 0 ) )
     {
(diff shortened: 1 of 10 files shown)


demos/projects/ESPRESSIF/esp32/components/sample-azure-iot/backoff_algorithm.c
```diff
@@ -46,7 +46,7 @@ BackoffAlgorithmStatus_t BackoffAlgorithm_GetNextBackoff( BackoffAlgorithmContex
 
     /* If BACKOFF_ALGORITHM_RETRY_FOREVER is set to 0, try forever. */
```
vaavva, line 47: In the L475 sample, this looks like this, and I'm wondering if that should be the case here as well instead of my change in the code to match the existing comment.
```
/* If maxRetryAttempts state of the context is set to the maximum, retry forever. */
if( ( pRetryContext->maxRetryAttempts == BACKOFF_ALGORITHM_RETRY_FOREVER ) ||
    ( pRetryContext->attemptsDone < pRetryContext->maxRetryAttempts ) )
{

File: demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c [hunk 2]
```diff
@@ -470,24 +494,34 @@ static void prvAzureDemoTask( void * pvParameters )
...
-         * packet. After sending disconnect, client must close the network
-         * connection. */
-        xResult = AzureIoTHubClient_Disconnect( &xAzureIoTHubClient );
-        configASSERT( xResult == eAzureIoTSuccess );
+            /* Send an MQTT Disconnect packet over the already connected TLS over
+             * TCP connection. There is no corresponding response for the disconnect
+             * packet. After sending disconnect, client must close the network
+             * connection. */
+            xResult = AzureIoTHubClient_Disconnect( &xAzureIoTHubClient );
+            configASSERT( xResult == eAzureIoTSuccess );
```
ericwol-msft, line 509: Assets should only be used on precondition that must be true. In this case xRestult can be a valid failure case. These should be just logging items. Example, what happens if you build release, these checks are not there anymore.

vaavva: I agree - I think we probably need to take a look at most of the asserts in the samples (~300) and see how many have valid failure cases that should have defined ways of handling them/logging and continuing. During standup, we had decided on the quick fix for this issue of logging on network issues and restarting the sample, but I can create a separate work item to take a look at the asserts in the samples and refactor them. Unfortunately for the ESP32, the assert function is in the esp library, so we can't easily override it like with some of the other devices.

CIPop (approved):
Thanks @vaavva !

PR #329, opened 2023-02-06
Recovery follow-up feedback

danewalton:
@jspaith gave some good feedback. Following up on those items.


vaavva (approved):

demos/projects/ESPRESSIF/az-ca-recovery/README.md
```diff
@@ -117,9 +117,11 @@ az account list-locations | ConvertFrom-Json | format-table -Property name
 # Create a resource group
 az group create --name '<name>' --location '<location>'
 
-# Deploy the resources
-# For the `resourcePrefix`, choose a unique, short name with only letter characters which
-# will be prepended to all of the deployed resources.
+# Deploy the resources. This may take a couple minutes.
+#  - `resourcePrefix`: choose a unique, short name with only lower-case letters which
+#                      will be prepended to all of the deployed resources.
```
jspaith, line 122: Nit "will be prepended" - consider "the script will prepend" to make it clear that the user has no further intervention to make after deciding this.

demos/projects/ESPRESSIF/az-ca-recovery/README.md
```diff
@@ -164,7 +166,11 @@ Select `Save` and make note of the Group Enrollment Key.
 
 ## Create a Derived Shared Access Key for Group Enrollment
 
-To use the custom allocation policy with the recovery Azure function, you have to create a derived SAS key from the group enrollment key. To create one, [follow the directions here](https://learn.microsoft.com/azure/iot-dps/concepts-symmetric-key-attestation?tabs=azure-cli#group-enrollments) and save the derived key to be used later. You may use a registration id of your choice.
+To use the custom allocation policy with the recovery Azure function, you have to create a derived SAS key from the group enrollment key. To create one, use the following Azure CLI command or [follow the directions here for other options](https://learn.microsoft.com/azure/iot-dps/concepts-symmetric-key-attestation?tabs=azure-cli#group-enrollments). You may use a registration id of your choice. Make sure to save this for the [RECOVERY] Derived Shared Access Key later. The `--key` parameter should be the group enrollment "Primary Key".
```
jspaith, line 169: Nit - recommend removing the `or [follow...]'.  I don't think there's much ambiguity in this path since it's the same cmd line, so we want to take this decision point away from users.  We could have something like at end of this "You can learn more about what this command is doing from [here](learn.ms.com/blah/blah)".

demos/projects/ESPRESSIF/az-ca-recovery/README.md
```diff
@@ -243,6 +251,16 @@ openssl x509 -in recovery-public-cert.pem -modulus -noout | sed s/Modulus=// | s
...
 ```
 
+For Powershell:
+
+```powershell
+# Modulus
+openssl x509 -in recovery-public-cert.pem -modulus -noout | ForEach-Object { $_ -replace 'Modulus=', '' } | ForEach-Object { $_ -replace '(..)', '0x$1, ' }
+
+# Exponent
+openssl x509 -in recovery-public-cert.pem --text -noout | Select-String Exponent |  ForEach-Object { $_ -replace 'Exponent: .*\((.*)\)', '$1' }
```
jspaith, line 261: This did not work on my `openssl` version on Windows 11 PowerShell.  When I used a single hyphen `-text -noout` it did.  

When I did this on bash (WSLv2 Ubuntu based distro specfiically) it seemed perfectly happy with both a single hyphen and double hyphen.

Is there a reason we need the double hyphen version?  If not, recommend we move to single hyphen.  Potentially points to me having an older openssl on Win (which I'll check in parallel) but I'm sure other customers could be in same boat I am so ideally we can make 1 cmd to make everyone work.

jspaith:
Thanks @danewalton for making all these changes!

jspaith (approved):

demos/projects/ESPRESSIF/az-ca-recovery/README.md
```diff
@@ -1,6 +1,6 @@
...
+Trust Bundle Recovery allows a mechanism for your device to recover connection to operational Azure resources should the CA certificates expire or otherwise become invalid. This requires an additional, separate "recovery" DPS instance and device credential, only to be used for recovery of the CA certificate. Should CA verification fail while connecting to operational IoT services, the device will connect to the recovery DPS instance, ignoring TLS CA cert validation, and will be returned a signed trust bundle recovery payload from DPS. This payload contains the complete and updated collection of CA certificates, asymmetrically signed by a root key created by the user. The device receives this payload, compares the signature using the saved public key, and then installs the certificates into the device's non-volatile storage (NVS). This permanent storage allows the new certificates to be loaded again should the device restart, removing the need for the device to use the recovery endpoint again. From that point on, the device can connect to the operational DPS endpoint and provisioned hub with full CA trust validation.
```
ericwol-msft, line 3:
```suggestion
Trust Bundle Recovery allows a mechanism for your device to recover connection to operational Azure resources should the TLS CA certificates expire or otherwise become invalid. This requires an additional, separate "recovery" DPS instance and device credential, only to be used for recovery of the CA certificates. Should CA verification fail while connecting to operational IoT services, the device should connect to the recovery DPS instance, ignoring the TLS CA cert validation, and will be returned a signed CA trust bundle recovery payload from the recovery DPS. This payload contains the complete and updated collection of the current CA certificates, asymmetrically signed by a root key created by the user. The device receives this payload, compares the signature using the saved public key, and then after successful validation, installs the certificates into the device's non-volatile storage (NVS). This permanent storage allows the new certificates to be loaded again should the device restart, removing the need for the device to reconnect to the recovery endpoint again. From that point on, the device can connect to the operational DPS endpoint and provisioned hub with full CA trust validation elabled.
```

ericwol-msft (approved):

PR #332, opened 2023-02-08
Add recovery in iothub case as well

danewalton:
Two scenarios tested:

## Scenario 1

Original. DPS and Hub both on Baltimore. Device has certs which don't include Baltimore. Failed connection to `[OPERATIONAL]` DPS, connect to recovery, then get trust bundle, etc. Works ✅.

## Scenario 2 (new)

DPS on Baltimore and Hub migrated to Digicert. Device has Baltimore. Device succeeds to `[OPERATIONAL]` DPS, but fails to `[OPERATIONAL]` Hub. Fall back to `[RECOVERY]` DPS to get trust bundle, etc. Then reprovision and connect to `[OPERATIONAL]` hub. Works ✅.

### DIFF
--- a/demos/sample_azure_iot_ca_recovery/sample_azure_iot_ca_recovery.c
+++ b/demos/sample_azure_iot_ca_recovery/sample_azure_iot_ca_recovery.c
@@ -349,19 +349,19 @@ static void runRecovery( NetworkCredentials_t * xNetworkCredentials,
     if( ( ulStatus = prvRunRecovery( xNetworkCredentials ) ) != 0 )
     {
         LogError( ( "Failed to run recovery error code = 0x%08x\r\n", ulStatus ) );
-        configASSERT( ulStatus == 0 );
+        configASSERT( false );
     }
     else if( ( ulStatus = prvSetupNetworkCredentials( xNetworkCredentials ) ) != 0 )
     {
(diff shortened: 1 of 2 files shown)


demos/sample_azure_iot_ca_recovery/sample_azure_iot_ca_recovery.c
```diff
@@ -334,6 +334,39 @@ static uint32_t prvSetupRecoveryNetworkCredentials( NetworkCredentials_t * pxNet
...
+    {
+        LogError( ( "Could not set network credentials\r\n" ) );
+        configASSERT( ulStatus == 0 );
+    }
+    else if( ( ulStatus = prvIoTHubInfoGet( xNetworkCredentials, pucIotHubHostname,
+                                            pulIothubHostnameLength, pucIotHubDeviceId,
+                                            pulIothubDeviceIdLength ) ) != 0 )
+    {
+        LogError( ( "Failed to run DPS after recovery!: error code = 0x%08x\r\n", ulStatus ) );
+        configASSERT( ulStatus == 0 );
```
vaavva, line 364: These asserts are guaranteed to fail if they are evaluated - should we have configAssert( false ) to be clearer about the flow?

vaavva (approved):

PR #333, opened 2023-02-10
Fix Windows sample for Release Builds

vaavva:
## Purpose

* Windows build would fail in release mode with `LNK2019 unresolved external symbol __except_handler4_common referenced in function __except_handler4 (MSVCRT.lib)` This is because the release build needs vcruntime.lib (https://stackoverflow.com/questions/31867018/unresolved-external-symbol-except-handler4-common-in-visual-studio-2015), which from this documentation (https://learn.microsoft.com/en-us/cpp/c-runtime-library/crt-library-features?view=msvc-170&viewFallbackFrom=vs-2019) shows that we need to use `/MD` instead of `/MT` in our `msvc_flags.cmake`
* This PR also adds building release in addition to debug in our pipeline
* This PR also adds the ability to manually trigger test runs on github.
    * How you'll be able to trigger: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
    * Docs on the change: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

build windows in both debug and release mode and make sure both work
Release:
```
cmake  -DVENDOR=PC -DCMAKE_GENERATOR_PLATFORM=Win32 -DCMAKE_BUILD_TYPE=Release -DBOARD=windows -Bbuild_windows .
cmake --build build_windows --config Release
```
Debug:
```
cmake  -DVENDOR=PC -DCMAKE_GENERATOR_PLATFORM=Win32 -DCMAKE_BUILD_TYPE=Debug -DBOARD=windows -Bbuild_windows .
cmake --build build_windows
```
## What to Check
* sample builds and runs as expected

### DIFF
--- a/.github/scripts/ci_tests_windows.ps1
+++ b/.github/scripts/ci_tests_windows.ps1
@@ -33,7 +33,9 @@ function sample_build
         [string] $board,
         [Parameter(Mandatory=$true, Position=2)]
         [string] $outdir,
-        [Parameter(Mandatory=$false, Position=3)]
+        [Parameter(Mandatory=$true, Position=3)]
+        [string] $buildType,
+        [Parameter(Mandatory=$false, Position=4)]
         [string] $additionalFlags
     )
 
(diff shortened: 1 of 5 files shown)


ewertons (approved):
Thanks for doing this, Val!

PR #334, opened 2023-02-10
[Asserts] Add panic handler and restart functions for all devices/samples

vaavva:
## Purpose

* First step to analyzing the existing asserts and having the device restart on failures in non-debug builds/loop for a debugger on debug builds

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/demos/common/sample_azure_common.h
+++ b/demos/common/sample_azure_common.h
@@ -12,15 +12,16 @@
 #include "demo_config.h"
 
 #if NDEBUG
-    #define panic_handler()                                                                           \
-    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\nRestarting device...\r\n", __FILE__, __func__, __LINE__ ); \
-    configRestartDevice()
+    #define panic_handler()                                                                                \
+    do { vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\nRestarting device...\r\n", __FILE__, __func__, __LINE__ ); \
+         democonfigRestartDevice();                                                                        \
+    } while( 0 )
 #else
-    #define panic_handler()                                                   \
-    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\n", __FILE__, __func__, __LINE__ ); \
-    vLoggingPrintf( "Looping to enable attaching a debugger\r\n" );           \
-    taskDISABLE_INTERRUPTS();                                                 \
-    for( ; ; ) {; }
+    #define panic_handler()                                                        \
+    do { vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\n", __FILE__, __func__, __LINE__ ); \
+         vLoggingPrintf( "Looping to enable attaching a debugger\r\n" );           \
+         for( ; ; ) {; }                                                           \
+    } while( 0 )
 #endif
 
 #endif /* ifndef SAMPLE_AZURE_COMMON_IF_H */
--- a/demos/common/sample_azure_common.h
+++ b/demos/common/sample_azure_common.h
@@ -1,0 +1,26 @@
+/* Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License. */
+
+/**
+ * @brief Contains helpers that are used in multiple samples
+ */
+
+#ifndef SAMPLE_AZURE_COMMON_IF_H
+#define SAMPLE_AZURE_COMMON_IF_H
+
+/* Demo Specific configs. */
+#include "demo_config.h"
+
+#if NDEBUG
+    #define panic_handler()                                                                           \
+    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\nRestarting device...\r\n", __FILE__, __func__, __LINE__ ); \
+    configRestartDevice()
+#else
+    #define panic_handler()                                                   \
+    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\n", __FILE__, __func__, __LINE__ ); \
+    vLoggingPrintf( "Looping to enable attaching a debugger\r\n" );           \
+    taskDISABLE_INTERRUPTS();                                                 \
+    for( ; ; ) {; }
+#endif
+
+#endif /* ifndef SAMPLE_AZURE_COMMON_IF_H */
--- a/demos/projects/ESPRESSIF/adu/config/demo_config.h
+++ b/demos/projects/ESPRESSIF/adu/config/demo_config.h
@@ -221,12 +221,12 @@
 /**
  * @brief Defines device restart function, used by the common sample modules.
  */
-#define configRestartDevice()    esp_restart()
+#define democonfigRestartDevice()    esp_restart()
 
 /**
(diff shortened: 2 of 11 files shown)


demos/common/sample_azure_common.h
```diff
@@ -0,0 +1,26 @@
...
+
+#if NDEBUG
+    #define panic_handler()                                                                           \
+    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\nRestarting device...\r\n", __FILE__, __func__, __LINE__ ); \
+    configRestartDevice()
+#else
+    #define panic_handler()                                                   \
+    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\n", __FILE__, __func__, __LINE__ ); \
+    vLoggingPrintf( "Looping to enable attaching a debugger\r\n" );           \
+    taskDISABLE_INTERRUPTS();                                                 \
```
CIPop, line 22: Why do we disable interrupts here? Our code shouldn't really ever need to create a critical section (instead we can use OS-level constructs such as mutexes, semaphores, etc).

demos/projects/ESPRESSIF/adu/config/demo_config.h
```diff
@@ -218,10 +218,15 @@
  */
 #define democonfigIOTHUB_PORT            8883
 
+/**
+ * @brief Defines device restart function, used by the common sample modules.
+ */
+#define configRestartDevice()    esp_restart()
```
CIPop, line 224: Why do we name these "config"? /cc @danewalton @ewertons 
IMO, they better fit a platform abstraction rather than configuration.

demos/projects/PC/linux/config/demo_config.h
```diff
@@ -220,17 +220,24 @@ extern void vLoggingPrintf( const char * pcFormatString,
...
  */
-#define democonfigIOTHUB_PORT                ( 8883 )
+#define democonfigIOTHUB_PORT            ( 8883 )
+
+#include <stdlib.h>
+
+/**
+ * @brief Defines device restart function, used by the common sample modules.
+ */
+#define configRestartDevice()    exit( EXIT_FAILURE )
```
CIPop, line 240: Nit: abort() will generate a core-dump that can be used to see how the application crashed.

demos/projects/PC/windows/config/demo_config.h
```diff
@@ -231,4 +231,11 @@ extern void vLoggingPrintf( const char * pcFormatString,
  */
 #define democonfigIOTHUB_PORT            ( 8883 )
 
+#include <stdlib.h>
+
+/**
+ * @brief Defines device restart function, used by the common sample modules.
+ */
+#define configRestartDevice()    exit( EXIT_FAILURE )
```
CIPop, line 239: Same as above.

CIPop (approved):
Left a few comments.

demos/projects/ESPRESSIF/adu/config/demo_config.h
vaavva: Agree - I was following the naming convention from the rest of the file, but I'm open to suggestions on the best name for this!

demos/common/sample_azure_common.h
vaavva: I was following FreeRTOS's existing configAssert implementation, but I'm open to other solutions that would enable attaching a debugger the best!

demos/common/sample_azure_common.h
```diff
@@ -0,0 +1,26 @@
...
+
+/* Demo Specific configs. */
+#include "demo_config.h"
+
+#if NDEBUG
+    #define panic_handler()                                                                           \
+    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\nRestarting device...\r\n", __FILE__, __func__, __LINE__ ); \
+    configRestartDevice()
+#else
+    #define panic_handler()                                                   \
```
danewalton, line 19: do while here too

demos/common/sample_azure_common.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef SAMPLE_AZURE_COMMON_IF_H
+#define SAMPLE_AZURE_COMMON_IF_H
+
+/* Demo Specific configs. */
+#include "demo_config.h"
+
+#if NDEBUG
+    #define panic_handler()                                                                           \
+    vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\nRestarting device...\r\n", __FILE__, __func__, __LINE__ ); \
+    configRestartDevice()
```
danewalton, line 17: consider using a `do {} while (0)` with multiple function calls in a macro: https://stackoverflow.com/questions/2687569/c-do-while0

```suggestion
    #define panic_handler()                                                                           \
    do { vLoggingPrintf( "[ERROR] [%s:%s:%d]\r\nRestarting device...\r\n", __FILE__, __func__, __LINE__ ); \
    configRestartDevice(); } while (0)
```

demos/projects/ESPRESSIF/adu/config/demo_config.h
danewalton, line 224:
```suggestion
#define democonfigRestartDevice()    esp_restart()
```

demos/projects/ESPRESSIF/adu/config/demo_config.h
```diff
@@ -218,10 +218,15 @@
...
+
 /**
  * @brief Defines configRAND32, used by the common sample modules.
  */
-#define configRAND32()    ( rand() / RAND_MAX )
+#define configRAND32()           ( rand() / RAND_MAX )
```
danewalton, line 229:
```suggestion
#define democonfigRAND32()           ( rand() / RAND_MAX )
```

demos/projects/ESPRESSIF/adu/config/demo_config.h
danewalton: I don't disagree. The naming convention is from FreeRTOS where you name #defines by prefixing with the filename. So in this case, `democonfigBLAH`.

We can surely create some kind of platform abstraction, but that would add a bunch more to this I think.

danewalton (approved):

demos/common/sample_azure_common.h
vaavva: removed the line to disable interrupts - let me know if there was something else you had in mind!

demos/projects/ESPRESSIF/adu/config/demo_config.h
vaavva: It looks like some of the samples define this in demo_config and some define it in FreeRTOSConfig, so I think that's why the prefix is just `config`. Probably best to leave it as is for now then?
https://github.com/Azure-Samples/iot-middleware-freertos-samples/search?q=configRAND32

PR #347, opened 2023-03-01
remove humidity sensor for crashing compiler

danewalton:
This sensor library is unused by ST but is causing a build break for our compilation (intermittently). Removing to prevent the random breaks.

https://github.com/Azure-Samples/iot-middleware-freertos-samples/actions/runs/4288570165/jobs/7470716793#step:11:624


danewalton:
Our getting started guide uses it along with the temp sensor. So can't be removed.

PR #351, opened 2023-03-03
Replace asserts with panic handler in main pnp sample file

vaavva:
## Purpose

* PR to replace assert usage in the main pnp sample file. In a debug build, there should be no functionality change, and in a release build, the device should restart on failures to try and recover (existing behavior is that the sample could get into weird broken states in a release build).

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
+++ b/demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
@@ -450,9 +450,9 @@ static void prvAzureDemoTask( void * pvParameters )
 
                 if( xResult == eAzureIoTErrorPublishFailed )
                 {
-                    LogError( ( "Error sending telemetry - restarting demo...\r\n" ) );
-                    /* force the demo to restart */
-                    break;
+                    LogError( ( "Error sending telemetry - restarting device...\r\n" ) );
+                    configRestartDevice();
+                    break; /* for PC samples */
                 }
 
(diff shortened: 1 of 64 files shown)


demos/sample_azure_iot_pnp/sample_azure_iot_pnp.c
```diff
@@ -420,77 +442,117 @@ static void prvAzureDemoTask( void * pvParameters )
...
+        else
+        {
+            /* Publish messages with QoS1, send and process Keep alive messages. */
+            for( ; ; )
             {
-                xResult = AzureIoTHubClient_SendTelemetry( &xAzureIoTHubClient,
-                                                           ucScratchBuffer, ulScratchBufferLength,
-                                                           NULL, eAzureIoTHubMessageQoS1, NULL );
-                configASSERT( xResult == eAzureIoTSuccess );
+                if( ( xResult = prvTelemetryLoop( ulScratchBufferLength ) ) != eAzureIoTSuccess )
```
danewalton, line 471: nit....this does more than just telmetry so consider something like `prvSendAndReceiveData` or something like that. Also doesn't technically loop so consider removing that?

danewalton (approved):

PR #354, opened 2023-03-09
prevent null ptr dereference

devotip:
## Purpose

* prevents null pointer dereference on error handling

```
[ ] Yes
[X] No
```

## Pull Request Type
fix

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/demos/projects/ESPRESSIF/adu/components/sample-azure-iot/transport_tls_esp32.c
+++ b/demos/projects/ESPRESSIF/adu/components/sample-azure-iot/transport_tls_esp32.c
@@ -134,6 +134,11 @@ TlsTransportStatus_t TLS_Socket_Connect( NetworkContext_t * pNetworkContext,
         esp_transport_ssl_set_client_key_data_der( pxEspTlsTransport->xTransport, (const char *) pNetworkCredentials->pucPrivateKey, pNetworkCredentials->xPrivateKeySize );
     }
 
+	if( pxEspTlsTransport == NULL )
+	{
+        xReturnStatus = eTLSTransportConnectFailure;
+	}
+	else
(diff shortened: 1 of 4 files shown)


devotip:
> 

@microsoft-github-policy-service agree

demos/projects/ESPRESSIF/adu/components/sample-azure-iot/transport_tls_esp32.c
```diff
@@ -134,6 +134,11 @@ TlsTransportStatus_t TLS_Socket_Connect( NetworkContext_t * pNetworkContext,
         esp_transport_ssl_set_client_key_data_der( pxEspTlsTransport->xTransport, (const char *) pNetworkCredentials->pucPrivateKey, pNetworkCredentials->xPrivateKeySize );
     }
 
+	if( pxEspTlsTransport == NULL )
```
RLeclair, line 137: There is already a NULL check for this variable beforehand in line 95. What is different in this case? Have you run into the null ptr dereference error?

devotip: Then the test in line 155 is redundant too

RLeclair: Unless there is a call in between 95 and 137 that can change `pxEspTlsTransport` to NULL (from what I see there are only calls setting variables in `pxEspTlsTransport` not the variable itself), the check you've added would not be hit. If that is not the case, could you point me to which function?

RLeclair:
Thank you so much for your contribution, we'll close this PR for now. If you would like to follow-up please free to re-open it.

Issue #338, opened 2023-02-21
PnP crashes after 64 minutes


Quinnvanderschaar2:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> esp32
pnp example

### Any log messages given by the failure
>

### Expected/desired behavior
> After one hour of sending  telemetry messages each minute,  the device crashes and gives the following error:
I (3947613) AZ IOT: -----------------send telemetry-----------------
I (3947633) AZ IOT: Successfully sent telemetry message
I (3947633) AZ IOT: Attempt to receive publish message from IoT Hub.
E (3947643) TRANSPORT_BASE: poll_read select error 104, errno = Connection reset by peer, fd = 54
E (3947643) esp_tls: Reading failed, errno= 119
E (3947643) MQTT: A single byte was not read from the transport: transportStatus=-1.
E (3947663) MQTT: Receiving incoming packet length failed. Status=MQTTRecvFailed
E (3947663) MQTT: Exiting process loop due to failure: ErrorStatus=MQTTRecvFailed
E (3947673) AZ IOT: AzureIoTMQTT_ProcessLoop failed: ProcessLoopDuration=30, MQTT error=0x00000004

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)
Linux Mint
### Versions
>I
ESP-IDF v4.4.3-159-g54dcccc1f6-dirty

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

danewalton:
Hi @Quinnvanderschaar2 
I assume you're using SAS keys for auth?

Quinnvanderschaar2:
Yes

danewalton:
I have this repro'd. Let me look into this.

danewalton:
Hey @Quinnvanderschaar2 
If you have a moment, can you try this branch here and see if it helps? This will let the device attempt a reconnect on process loop failure. In MQTT 3, a token expiration requires a disconnect and reconnect, which this does. Also to note for you and others, the extra ~5 minutes above the default 60 minutes is to account for potential clock drift.

[...]

PR #372, opened 2023-06-08
Fix esp32 sample memory leaks; donot reset if internet disconnects

ewertons:
## Purpose

* ...
Fix memory leaks in ESP32 sample in between sample main task loops.
Add code to avoid resetting ESP32 board if internet connection drops.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
Run the ESP32 sample on a real device. Observe the free RAM memory after each loop of the main task.

## What to Check
You can use this trace statement for verifying the available RAM memory:

```c
        LogInfo( ( "[Debug][loop-end] Free heap (bytes): %lu, %lu (int), %lu (min)",
            esp_get_free_heap_size( ),
            esp_get_free_internal_heap_size( ),
            esp_get_minimum_free_heap_size( ) ) );
```

## Other Information
Spearmint contains 0.5% menthol. Peppermint contains 40%! How to differentiate both? Eat it and feel the freshness.


ewertons:
May resolve https://github.com/Azure-Samples/iot-middleware-freertos-samples/issues/338, https://github.com/Azure/azure-iot-middleware-freertos/issues/234

ericwol-msft (approved):

demos/projects/ESPRESSIF/adu/main/azure_iot_freertos_esp32_main.c
```diff
@@ -76,6 +76,8 @@ static bool xTimeInitialized = false;
 static xSemaphoreHandle xSemphGetIpAddrs;
 static esp_ip4_addr_t xIpAddress;
 
+static bool s_is_sample_connected_to_internet = false;
```
ericwol-msft, line 79:
```suggestion
static bool s_is_connected_to_internet = false;
```

demos/projects/ESPRESSIF/esp32/components/sample-azure-iot/transport_tls_esp32.c
```diff
@@ -269,73 +269,83 @@ TlsTransportStatus_t TLS_Socket_Connect( NetworkContext_t * pNetworkContext,
...
-    // Create a transport list into which we put the transport.
-    pxEspTlsTransport->xTransportList = esp_transport_list_init();
-    pxEspTlsTransport->xTransport = esp_transport_ssl_init( );
-    pxEspTlsTransport->ulReceiveTimeoutMs = ulReceiveTimeoutMs;
-    pxEspTlsTransport->ulSendTimeoutMs = ulSendTimeoutMs;
-
-    esp_transport_ssl_enable_global_ca_store(pxEspTlsTransport->xTransport);
+        if(pxEspTlsTransport == NULL)
+        {
+        return eTLSTransportInsufficientMemory;
```
ericwol-msft, line 284: nit...TAB indent

demos/projects/NXP/mimxrt1060/main.c
```diff
@@ -241,6 +244,12 @@ void CRYPTO_InitHardware( void )
 }
 /*-----------------------------------------------------------*/
 
+bool xAzureSample_IsConnectedToInternet()
+{
+    return true;
```
vaavva, line 249: For clarity, can you add a comment saying this isn't implemented for now in the samples where it isn't?

demos/sample_azure_iot/sample_azure_iot.c
```diff
@@ -353,145 +356,154 @@ static void prvAzureDemoTask( void * pvParameters )
...
-                                                                    ucScratchBuffer, ulScratchBufferLength,
-                                                                    NULL );
+                                                  sampleazureiotMESSAGE, lPublishCount );
+                xResult = AzureIoTHubClient_SendTelemetry( &xAzureIoTHubClient,
+                                                           ucScratchBuffer, ulScratchBufferLength,
+                                                           &xPropertyBag, eAzureIoTHubMessageQoS1, NULL );
+                configASSERT( xResult == eAzureIoTSuccess );
+
+                LogInfo( ( "Attempt to receive publish message from IoT Hub.\r\n" ) );
+                xResult = AzureIoTHubClient_ProcessLoop( &xAzureIoTHubClient,
```
vaavva, line 460: If we lose connectivity during one of these steps, will that cause the configAssert to crash the sample instead of getting caught by the xAzureSample_IsConnectedToInternet in the loop check?

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
```diff
@@ -918,33 +888,72 @@ static void prvAzureDemoTask( void * pvParameters )
...
-            vTaskDelay( sampleazureiotDELAY_BETWEEN_PUBLISHES_TICKS );
-        }
+                /* Leave Connection Idle for some time. */
+                LogInfo( ( "Keeping Connection Idle..." ) );
+                vTaskDelay( sampleazureiotDELAY_BETWEEN_PUBLISHES_TICKS );
+            }
 
-        xResult = AzureIoTHubClient_UnsubscribeProperties( &xAzureIoTHubClient );
-        configASSERT( xResult == eAzureIoTSuccess );
+            if( xAzureSample_IsConnectedToInternet() )
```
vaavva, line 938: Before, we couldn't get to this point because of the for (; ; ) - this will change the behavior of the sample now that this code is reachable

vaavva, line 938: Also, should this be outside of the for( ; xAzureSample_IsConnectedToInternet(); ) so that everything gets shut down properly if the connection is dropped?

vaavva (approved):

RLeclair (approved):

demos/projects/ESPRESSIF/adu/main/azure_iot_freertos_esp32_main.c
ewertons: Taken.

demos/projects/NXP/mimxrt1060/main.c
ewertons: Makes sense. Done.

demos/sample_azure_iot/sample_azure_iot.c
ewertons: Yes it will. The internet connection checks are for now just for some key points of the code. 
More test will be done later to try to extend the check to all potential I/O failure points.

demos/sample_azure_iot_adu/sample_azure_iot_adu.c
ewertons: Correct. In reality this could still be unreachable, since to get here there must NOT be internet connection. And then if the check again confirms there is still no internet connection, the graceful disconnection commands will not be run. To improve this we would need to have further refactoring.

wreyford:
Thx a lot everyone. This memory leak that is sorted by esp_tls_free_global_ca_store(); has helped resolve a major memory leak in our project. 

I still have to find a way to get around the IoTHub disconnecting me after about 1 hour, causing an assert and resetting our device. (And that is with a valid internet connection).

rtheil-growlink:
> I still have to find a way to get around the IoTHub disconnecting me after about 1 hour, causing an assert and resetting our device. (And that is with a valid internet connection).

Please see this issue for some additional info: https://github.com/Azure/azure-iot-middleware-freertos/issues/234

wreyford:
Wow, thanks a lot @rtheil-growlink . Your solution mentioned above has been implemented and appears to be working just fine. 

I wonder why it has not been included in the base code examples.

